PR #2, opened 2025-09-02
[GUHEALTH-369] Add fault injection functionality

ambermingxin:


cmd/gpuhealth/command/command.go
```diff
@@ -188,6 +189,34 @@ func App() *cli.App {
...
+				},
+				&cli.StringFlag{
+					Name:  "fault-message",
+					Usage: "message to inject into the component",
+				},
+				&cli.StringFlag{
+					Name:  "event-type",
+					Usage: "type of the event to inject into the component",
+				},
+				&cli.StringFlag{
```
vinodba: why is server address specification here? Also is the usage that the agent daemon is running and user runs this command line to add the fault injection?

cmd/gpuhealth/command/command.go
```diff
@@ -188,6 +189,34 @@ func App() *cli.App {
...
+			Action: cmdinject.Command,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:     "component,c",
+					Usage:    "component name to inject fault into (required)",
+					Required: true,
+				},
+				&cli.StringFlag{
+					Name:  "fault-type",
+					Usage: "fault type to inject into the component (component-error or event)",
```
vinodba: code has type kernel-message. do we need to add that here?

cmd/gpuhealth/command/command.go
ambermingxin: this would be default to the current mock health endpoint if not set. i could remove it too but anyways we need to change this mock endpoint to the real one at some point

ambermingxin: > is the usage that the agent daemon is running and user runs this command line to add the fault injection?

yes. exactly what i demo last friday

cmd/gpuhealth/command/command.go
ambermingxin: yes. kmsg is another type. i will add to the usage description here too.

PR #29, opened 2025-10-29
[GPUHEALTH-577] Agent Enrollment uses proper endpoint

mukilsh:
User enters

sudo ./bin/gpuhealth enroll --endpoint https://data.us-e/
ast-1.dev.gpuhealth.nvidia.com --token [token]

Rather than 

sudo ./bin/gpuhealth enroll --endpoint https://data.us-e/
ast-1.dev.gpuhealth.nvidia.com/api/v1/health --token [token]

### DIFF
--- a/cmd/gpuhealth/enroll.go
+++ b/cmd/gpuhealth/enroll.go
@@ -43,8 +43,14 @@ func enrollCommand(cliContext *cli.Context) error {
 		return fmt.Errorf("enrollment endpoint must use HTTPS, got %s", baseURL.Scheme)
 	}
 
-	// Construct enroll endpoint first
-	enrollEndpoint, err := url.JoinPath(baseURL.String(), "enroll")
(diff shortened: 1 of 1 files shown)


jingxiang-z (commented):
One comment about naming convention is that /api/v1/health is usually used for endpoint health probe and is a get method. Could we make the metrics endpoint as /api/v1/metrics, and logs endpoint /api/v1/logs, etc?

mukilsh:
> One comment about naming convention is that /api/v1/health is usually used for endpoint health probe and is a get method. Could we make the metrics endpoint as /api/v1/metrics, and logs endpoint /api/v1/logs, etc?

We can - will require changes on ELB, envoy, backend side as well. Lets maybe do all that in another round of fixes afterwards?

jingxiang-z (approved):

PR #31, opened 2025-11-03
[GPUHEALTH-721] feat: Add DCGM components to gpuhealth agent

jingxiang-z:


cmd/gpuhealth/root.go
ambermingxin, line 1: I understand we need to disable a few components e.g.nfs. How does this relate to dcgm?

deployments/packages/systemd/gpuhealthd.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=GPU Health Daemon
-After=network.target local-fs.target
+Wants=nvidia-dcgm.service
```
ambermingxin, line 3: What's the behavior if the customer doesn't have dcgm setup?

internal/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+		},
+		{
+			Name:             componentspci.Name,
+			InitFunc:         componentspci.New,
+			EnabledByDefault: true,
+		},
+	}
+}
+
+// GetEnabledComponents returns only the components that should be enabled by default
```
ambermingxin, line 222: what's the default enabled components? where are the configurations to enable/disable different components on agent side?

internal/registry/registry.go
```diff
@@ -0,0 +1,241 @@
...
+	var enabled []Component
+	for _, c := range All() {
+		if c.EnabledByDefault {
+			enabled = append(enabled, c)
+		}
+	}
+	return enabled
+}
+
+// GetComponent returns a component by name, or nil if not found
```
ambermingxin, line 233: Where is the reference/caller of this function

cmd/gpuhealth/root.go
jingxiang-z: This PR also address the component deduplication as documented [here](https://docs.google.com/spreadsheets/d/1R3bPvcA6KrJ8Hj1oaFNCujJevUYaS_Y99Znfb6McS24/edit?gid=1754453754#gid=1754453754).

deployments/packages/systemd/gpuhealthd.service
jingxiang-z: Then it just skips the dcgm. The Wants is a soft requirement that can be bypassed.

internal/registry/registry.go
jingxiang-z: I created this registry to disable some GPUd components from code that we don't want. The reamaining are also default to be enabled. The user has the choice to disable some of them from the config /etc/default/gpuhealth

internal/registry/registry.go
jingxiang-z: Both the server and the scan will call this function to initialize the compoenents defined here.

ambermingxin (approved):

PR #32, opened 2025-11-04
[GPUHEALTH-748] fix: gpuhealth database purge fix

jingxiang-z:


ambermingxin (approved):
Future work - combine the purge interval + retention period together to have a just one logic to keep the data in db instead of running goroutine to check db every some time.

PR #47, opened 2025-11-26
[GPUHEALTH-845] feat: call attestation sdk CLI with expected failure

ambermingxin:


internal/attestation/attestation.go
```diff
@@ -399,29 +400,46 @@ func (m *Manager) getMachineIdWithVBIOS() (string, []string, error) {
...
 func (m *Manager) getEvidences(nonce string, vbiosVersions []string) (*AttestationSDKResponse, error) {
-	log.Logger.Debugw("Getting evidences with parameters",
-		"nonce", nonce,
-		"vbios_versions", vbiosVersions,
-		"gpu_count", len(vbiosVersions))
+	log.Logger.Infow("Calling attestation SDK CLI", "nonce", nonce)
+
+	// Execute nvattest command
+	// Set timeout to prevent hanging
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
mukilsh: Can we use manager's context here so this respects cancellation of manager? If manager goes down, we probably can cancel this SDK call as well

internal/attestation/attestation.go
```diff
@@ -399,29 +400,46 @@ func (m *Manager) getMachineIdWithVBIOS() (string, []string, error) {
 }
 
 func (m *Manager) getEvidences(nonce string, vbiosVersions []string) (*AttestationSDKResponse, error) {
```
mukilsh, line 402: I think we don't use vbiosVersions here, maybe we can remove it from the function signature?

ambermingxin: I keep it here because we will use it after their sdk is implementation complete based on their [design doc](https://docs.google.com/document/d/18unXW48Km0YLJ39PQIbSlE5ODEDOXwolA_yoE04mWWE/edit?tab=t.0#heading=h.8qvrjsqqnoyw). you can see that fw version is listed as an input parameter.

mukilsh (approved):
LGTM

PR #51, opened 2025-12-04
[GPUHEALTH-885] fix: machine uuid from database

jingxiang-z:


internal/server/server.go
```diff
@@ -98,28 +99,36 @@ func initializeDatabases(ctx context.Context, config *config.Config) (*sql.DB, *
...
-		machineID = pkghost.MachineID()
-		if machineID == "" {
-			// Fallback to dynamic lookup if not cached
-			var err error
-			machineID, err = pkghost.GetMachineID(ctx)
-			if err != nil {
-				return "", fmt.Errorf("failed to get system machine ID: %w", err)
-			}
+		// First, try to get hardware UUID from dmidecode
+		machineID, err = pkghost.GetDmidecodeUUID(ctx)
```
ooolafhu, line 114: As this function returns a string, is it guaranteed to be either an empty string or a standard UUID format?

ooolafhu (approved):
Overall LGTM. Thanks a lot for making the changes.

internal/server/server.go
jingxiang-z: Yes.

PR #53, opened 2025-12-08
[GPUHEALTH-831] fix: private IP missing in gpuhealth machine-info

jingxiang-z:


cmd/gpuhealth/machine_info.go
```diff
@@ -81,6 +81,19 @@ func machineInfoCommand(cliContext *cli.Context) error {
 	if providerInfo == nil {
 		fmt.Printf("%s failed to find provider (%v)\n", cmdutil.WarningSign, err)
 	} else {
+		if providerInfo.PrivateIP == "" {
+			if machineInfo != nil && machineInfo.NICInfo != nil {
+				for _, iface := range machineInfo.NICInfo.PrivateIPInterfaces {
+					if iface.IP == "" {
+						continue
```
ambermingxin: There are 4 levels of nesting added to the main flow. why not add a helper function to reduce complexity. also it probably belongs closer to the data definition in `internal/machineinfo`.

ambermingxin (approved):

PR #69, opened 2026-01-07
[GPUHEALTH-1027] build: add skyhook package for install gpuhealth agent

jingxiang-z:


deployments/skyhooks/skyhook_dir/install_gpuhealth.sh
```diff
@@ -0,0 +1,142 @@
...
+
+require_env GPUHEALTH_VERSION
+require_env GPUHEALTH_ENDPOINT
+require_env GPUHEALTH_TOKEN
+require_env SKYHOOK_DIR
+
+# figure out the distro by sourcing the os-release file
+source /etc/os-release
+case "${ID:-}" in
+    ubuntu*)
```
ebalduf, line 46: We should probably change this to ubuntu* |  debian* )

ebalduf (approved):
We should probably change the case statement to support real Debian as well as Ubuntu, but otherwise this looks good.

ebalduf:
For now, lets not do the Debian change, since it will cause more testing.  I'll move that request to a JIRA ticket to add the support later.

PR #70, opened 2026-01-13
[GPUHEALTH-1000] Send Agent Config

mukilsh:


jingxiang-z:
There is a field called agentConfig.totalComponents in createOTLPResource function. Can we extend this to export other agent configuration like agentConfig.exportInterval?

jingxiang-z (approved):
LGTM

PR #72, opened 2026-01-14
Remove attestation from otel log

ambermingxin:


mukilsh (approved):
Logs part looks good. Seems like might be some formatting issues. But overall looks good!

PR #76, opened 2026-01-23
[GPUHEALTH-1168] feat: default bind to localhost for agent endpoint

jingxiang-z:


cmd/gpuhealth/root.go
```diff
@@ -87,7 +87,7 @@ func App() *cli.App {
 				&cli.StringFlag{
 					Name:  "listen-address",
 					Usage: "set the listen address",
-					Value: fmt.Sprintf("0.0.0.0:%d", config.DefaultHealthPort),
+					Value: fmt.Sprintf("127.0.0.1:%d", config.DefaultHealthPort),
```
rsampaio: instead of hardcoding the address, we should move to a constant like `DefaultHealthPort`, maybe something like `DefaultLIstenAddress`

internal/config/default.go
```diff
@@ -50,7 +50,7 @@ func Default(ctx context.Context, opts ...OpOption) (*Config, error) {
 
 	cfg := &Config{
 		APIVersion:      DefaultAPIVersion,
-		Address:         fmt.Sprintf(":%d", DefaultHealthPort),
+		Address:         fmt.Sprintf("127.0.0.1:%d", DefaultHealthPort),
```
rsampaio: Similar to the previous comment about moving to a constant

rsampaio: I see that we check if the listenAddress argument is empty if not we use that as `cfg.Address` then in the server.go file we use `cfg.Address` to listen to this address, in the inject.go file we are also hardcoding localhost and I wonder if we should refer to the `cfg.Address` when we use that value to inject

rsampaio:
Left a few comments about a constant to represent the address and some hardcoded values, and I would like to see a section on the documentation about prometheus and other monitoring tools to start the agent with the exposed argument!

jingxiang-z:
@codex review

rsampaio (approved):
Nice, this looks great 👍

PR #82, opened 2026-01-29
[gpuhealth-1197] Restrict access to /inject-fault endpoint and add option to explicit enable it

rsampaio:
A security analysis pointed out a possible security risk with a writable endpoint that injects faults (including a kernel message) using the agent's API

This change restrict the endpoint to localhost using `ClientIP()` and add an option to explicit enable this endpoint for testing. The endpoint will be disabled by default and documentation is updated to reflect the change and explain how this endpoint is used

### DIFF
--- a/cmd/gpuhealth/run.go
+++ b/cmd/gpuhealth/run.go
@@ -254,11 +254,6 @@ func runCommand(cliContext *cli.Context) error {
 	}
 	log.Logger.Infow("health exporter configuration", "cfg", cfg.HealthExporter)
 
-	// Apply environment variable overrides for feature flags
-	if err := setBoolFromEnv("GPUHEALTH_ENABLE_FAULT_INJECTION", &cfg.EnableFaultInjection, "set fault injection enabled from env", "enable_fault_injection"); err != nil {
-		return err
-	}
-
 	if listenAddress != "" {
 		cfg.Address = listenAddress
 	}
--- a/cmd/gpuhealth/run.go
+++ b/cmd/gpuhealth/run.go
@@ -194,6 +194,7 @@ func runCommand(cliContext *cli.Context) error {
 	gpuCount := cliContext.Int("gpu-count")
 	infinibandExpectedPortStates := cliContext.String("infiniband-expected-port-states")
 	enableDCGMPolicy := cliContext.Bool("enable-dcgm-policy")
+	enableFaultInjection := cliContext.Bool("enable-fault-injection")
 
 	// GPU Health Exporter configuration
 	offlineMode := cliContext.Bool("offline-mode")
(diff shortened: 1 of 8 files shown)


jingxiang-z:
@rsampaio Could you follow the PR standard here to signoff your commit. https://github.com/NVIDIA/gpuhealth/blob/main/CONTRIBUTING.md#pull-request-process. Also could you fix the unit test and add some unit tests to the PR?

internal/server/handlers_inject_fault.go
```diff
@@ -41,8 +42,22 @@ const URLPathInjectFault = "/inject-fault"
 // @Failure 500 {object} map[string]interface{} "Internal server error"
 // @Router /inject-fault [post]
 func (s *Server) injectFault(c *gin.Context) {
+	// Security check: verify request is from localhost
+	remoteHost, _, err := net.SplitHostPort(c.Request.RemoteAddr)
```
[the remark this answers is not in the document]
rsampaio: giving that we are disabling this endpoint by default, I believe it is ok to filter by client-ip but open to feedback here!

cmd/gpuhealth/run.go
```diff
@@ -253,6 +254,11 @@ func runCommand(cliContext *cli.Context) error {
 	}
 	log.Logger.Infow("health exporter configuration", "cfg", cfg.HealthExporter)
 
+	// Apply environment variable overrides for feature flags
+	if err := setBoolFromEnv("GPUHEALTH_ENABLE_FAULT_INJECTION", &cfg.EnableFaultInjection, "set fault injection enabled from env", "enable_fault_injection"); err != nil {
+		return err
+	}
+
```
jingxiang-z: Does this mean, it could be controlled by both the cli flag and the env var? What's the behavior if I set enable fault injection in the cli flag while have env var set to false?

jingxiang-z: Is there a reason that we want this to be controlled by both env var and cli flag, could it be one of them?

rsampaio: the CLI overrides the env var, my intent was to make it easier to quickly enable without changing config or adding extra parameters, I can favor one of them if we don't expect users to enable this endpoint often or at all

jingxiang-z: Yes, so usually I use cli flag if I don't expect users to enable this endpoint, for my own testing, I just run the binary with specific flag.

rsampaio: gotcha, yes that makes sense, I will remove the env var 👍

jingxiang-z (approved):
I prefer to only use cli flag to control this fault injection endpoint enablement so we have single source of truth. But currently it also works and approved.

PR #89, opened 2026-02-09
fix: unenroll when enroll flag is disabled

jingxiang-z:


deployments/helm/gpuhealth-agent/templates/daemonset.yaml
```diff
@@ -67,6 +67,21 @@ spec:
...
+          image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+            - /bin/bash
+            - -c
+          args:
+            - >
+              /usr/bin/gpuhealth unenroll || true
+          securityContext:
+            {{- toYaml .Values.enroll.securityContext | nindent 12 }}
```
ambermingxin: what's the purpose of `nindent 12`?

jingxiang-z: indents every line by 12 spaces so it nests under securityContext: correctly becasue securityContext could be multiline from the values file

ambermingxin: Hardcoding `nindent 12 `assumes `securityContext:` always starts at a specific column, which seems error-prone. maybe nit though so will give a go.

ambermingxin (approved):

PR #91, opened 2026-02-10
fix: dev mount directly to avoid nvidiactl issue writing file

jingxiang-z:


rsampaio (approved):
For future references, we already mount /dev as RW in the main container for the attestation and this fixes the problem with the init container 👍

PR #95, opened 2026-02-11
chore: add agent-artifact team to NGC for chart and image

jingxiang-z:


.github/workflows/release.yml
```diff
@@ -95,8 +95,8 @@ jobs:
       # - NVCR_API_KEY: NGC API key for nvcr.io (docker username is the literal '$oauthtoken')
       #
       # Image is pushed to:
-      # - nvcr.io/0753215517602916/gpuhealth-skyhook:<version>
-      # - nvcr.io/0753215517602916/gpuhealth-skyhook:latest (stable releases only)
+      # - nvcr.io/0753215517602916/agent-artifact/gpuhealth-skyhook:<version>
```
ambermingxin, line 98: we have the stg env whose org name is 0760200452382209 and we normally put the resources in both place. do we only use the prod one here?

jingxiang-z: Yes, we only need the prod registry. This is to share the image with external user

ambermingxin (approved):

PR #103, opened 2026-02-23
feat: add k8s env var in helm install

ambermingxin:


jingxiang-z:
Actually, can we make the change to parse the bool in the gpud code
`if os.Getenv("IS_RUNTIME_K8S") != "" `
Worry about if customer set this to false, they are expecting a different behavior.

PR #105, opened 2026-02-24
[GPUHEALTH-1361] Add Jitter to Intial OTLP Push to avoid thundering herd

mukilsh:


rsampaio:
@mukilsh is this PR still relevant? can we close it?

mukilsh:
> @mukilsh is this PR still relevant? can we close it?

Hmm. Not a requirement, but I actually think it would still be useful for sending to the gateway (to not overwhelm it). What do you think?

rsampaio:
@mukilsh I agree, Codex has a relevant comment about the request timeout being 30s and the jitter being 1 minute by default, the context may close before the jitter is introduced, besides that I think this is a good idea!

mukilsh:
@rsampaio got it - let me take a peek and get this MR ready again, I think we can still merge it. No urgency though, doesn't need to be in this immediate release

PR #111, opened 2026-03-03
[GPUHEALTH-1422] GPUHealth changes to vendor dependency libraries from GPUd

rsampaio:
## Summary

This PR replaces the external `github.com/leptonai/gpud` dependency with a **vendored, pruned copy** under `third_party/fleet-intelligence-sdk`, renames the module to `github.com/NVIDIA/fleet-intelligence-sdk`, and updates the Dockerfile so the image build succeeds.

---

### 1. Vendored dependency
- Switched from relying on a `replace` to a remote gpud (e.g. gitlab) to an **in-repo copy**.
- Added a pruned copy of the gpud packages actually used by the agent into `third_party/`, so the project no longer depends on an external gpud repo at build time.

### 2. Module rename to fleet-intelligence-sdk
- Renamed the vendored module to **`github.com/NVIDIA/fleet-intelligence-sdk`**.
- Updated the vendored tree under `third_party/fleet-intelligence-sdk/` to use that module path (its own `go.mod` and internal imports).
- Updated **`go.mod`** to `require github.com/NVIDIA/fleet-intelligence-sdk` with `replace ... => ./third_party/fleet-intelligence-sdk`.
- Updated **all agent imports** from `github.com/leptonai/gpud` to `github.com/NVIDIA/fleet-intelligence-sdk` across `cmd/` and `internal/` (registry, server, scan, exporter, config, attestation, machineinfo, enrollment, etc.).

### 3. Go version and Docker build fix
- Pinned the build to **Go 1.25.5** (in `go.mod` and Dockerfile base image).
- **Dockerfile:** `go mod download` runs before `COPY . .`, but the replace points to `./third_party/fleet-intelligence-sdk`, which was not in the image yet. Added an explicit **`COPY third_party/fleet-intelligence-sdk`** before `RUN go mod download` so the replace is satisfied and module download succeeds.
- Aligned any remaining imports to **fleet-intelligence-sdk** so the container build completes.

---

## Testing

- **`make docker-build`** completes and produces the `fleet-intelligence-agent:dev` image.
- Local build and tests: `make all` (or `make fleetint`), `make test`, `make lint`.

---

## Notes

- The vendored SDK is a **subset** of gpud (only the packages used by the agent).


go.mod
```diff
@@ -1,15 +1,13 @@
 module github.com/NVIDIA/fleet-intelligence-agent
 
-go 1.24.13
-
-replace github.com/leptonai/gpud => gitlab-master.nvidia.com/gpu-health/gpud v0.9.2-0.20260303003844-3bce642cf157
+go 1.25.5
```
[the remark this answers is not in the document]
jingxiang-z: Why do we upgrade the go version to 1.25? If no specific reason, can we keep the go version as 1.24.13? I remembered that before some CI will fail for the go 1.25 because some package doesn't support 1.25 yet.

jingxiang-z:
Can we make sure the latest change from Amber is there in the sdk?

jingxiang-z:
fleet-intelligence-sdk is a good name and maybe we can name it in the future for our health check repo. But because right now we are vendoring gpud, do you think it makes sense just to call this third_party/gpud, so that users know we have dependencies on lepton gpud?

third_party/fleet-intelligence-sdk/cmd/gpud/command/command.go
```diff
@@ -0,0 +1,770 @@
+package command
```
jingxiang-z, line 1: Do we need any packages from the gpud cmd? Those are gpud's binary. If not, maybe we can remove the entire cmd folder here.

go.mod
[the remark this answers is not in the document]
rsampaio: I believe it happened when I ran the script to prune the vendor dependencies, I will move back to 1.24.13

third_party/fleet-intelligence-sdk/cmd/gpud/command/command.go
rsampaio: at first the scan.go in our project was importing the cmdcommon package but I noticed we had the same constants in our source tree, I replaced that in scan.go but a few packages inside the vendor library still uses the same cmdcommon for messaging, I believe we can keep that vendored for now and refactor in the future.

rsampaio:
@jingxiang-z for the name, I feel like we would need less replace if we keep the name as we intend to opensource, in the future when the standalone library is public we just need to change the go.mod file and run `go mod tidy`, I would prefer to avoid big renames which would generate huge PRs

.github/workflows/release.yml
```diff
@@ -6,10 +6,11 @@ on:
       - 'v*'  # Triggers on version tags like v1.0.0, v2.1.3, etc.
 
 env:
-  GO_VERSION: '1.24.11'
+  GO_VERSION: '1.24.13'
   GOPRIVATE: gitlab-master.nvidia.com/*
   GONOPROXY: gitlab-master.nvidia.com/*
   GONOSUMDB: gitlab-master.nvidia.com/*
```
jingxiang-z: Nit: I think we can remove these since we vendor the gpud in this repo. But we can do this cleanup in a future PR.

.golangci.yml
```diff
@@ -1,6 +1,6 @@
 version: "2"
 run:
-  go: 1.24.9
+  go: 1.25.5
```
jingxiang-z: Should this be 1.24.13?

jingxiang-z (approved):
Look good to me. Please squash your commit into one before merging.

.github/workflows/release.yml
rsampaio: yes I agree, the local docker testing I added here passes without this, this is a good cleanup 👍

.golangci.yml
rsampaio: Ops, forgot to move this one back, let me change that! good catch!

PR #119, opened 2026-03-05
Add missing dependencies for vendor sdk tests and adjust dockerfile

rsampaio:
Tests from the vendor SDK were failing because of missing modules not initially imported and discrepancy between the go.mod upstream and the vendor directory, right now the majority of tests pass on a docker container and the failing ones are hardware related or container-host related.


jingxiang-z (approved):
LGTM

PR #132, opened 2026-03-12
[GPUHEALTH-1520] feat: add gpu index to machine info

ambermingxin:


internal/machineinfo/machineinfo.go
```diff
@@ -73,8 +73,29 @@ type MachineInfo struct {
...
-// GetMachineInfo retrieves machine info and customizes it for Fleet Intelligence
-func GetMachineInfo(nvmlInstance nvidianvml.Instance) (*MachineInfo, error) {
+// MachineInfoOption configures GetMachineInfo behavior.
+type MachineInfoOption func(*machineInfoOpts)
+
+type machineInfoOpts struct {
+	// dcgmGPUIndexes overrides NVML-sourced GPUIndex values with DCGM device IDs.
+	// Key: GPU UUID, Value: DCGM device ID as string (e.g. "0", "1").
+	dcgmGPUIndexes map[string]string
+}
```
jingxiang-z, line 83: Instead of override, can we have a separate index called DCGM GPU Index, separate from NVML GPU Index. Would like customer to have more information with this index.

ambermingxin: The NVML index is not surfaced independently. We don't expose the NVML GPU index as a standalone label or field in any customer-facing output (otlp, cli, or api). Introducing a second index would raise "which one do I use?" questions without a practical use case, since the only consumer of GPUIndex is the OTLP "gpu" label - which must match what DCGM emits. Therefore I feel like two indices increases ambiguity, not clarity. If customers see both an NVML index and a DCGM index in machine info but only one "gpu" label in metrics, it creates confusion about which value corresponds to their telemetry data. A single, authoritative index that matches the metrics is more reasonable to me.

mukilsh (approved):
LGTM

internal/machineinfo/machineinfo.go
ambermingxin: i will resolve this if there is no followup comment

PR #135, opened 2026-03-13
chore: add goheader linter to enforce license headers

rsampaio:
Adds the goheader linter to golangci-lint configuration to ensure all new Go files include the required SPDX Apache 2.0 license header.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,7 +13,28 @@ linters:
     - staticcheck
     - unconvert
     - unused
+    - goheader
   settings:
+    goheader:
+      values:
+        regexp:
+          YEAR: \d{4}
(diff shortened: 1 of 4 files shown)


.golangci.yml
```diff
@@ -13,7 +13,28 @@ linters:
     - staticcheck
     - unconvert
     - unused
+    - goheader
```
jingxiang-z, line 16: Does this golangci-lint run on the third-party directory?

rsampaio: there is a specific check for third_party that check for either a Copyright or SPDX header and I added that to the github action @jingxiang-z

jingxiang-z (approved):

PR #139, opened 2026-03-19
docs: improve install instructions with curl and arch detection

jingxiang-z:
## Summary

- Replace manual download steps in `install-deb.md` and `install-rpm.md` with `curl` commands that auto-detect the latest release version via the GitHub API
- Use `dpkg --print-architecture` for DEB arch detection (`amd64`/`arm64` matches DEB filename convention)
- Use `uname -m` for RPM arch detection (`x86_64`/`aarch64` matches RPM filename convention)
- Replace `wget` with `curl -fLO` in the CUDA keyring setup section of `install-deb.md` for consistency

## Test plan

- [x] Verify DEB install command on Ubuntu x86_64
- [x] Verify DEB install command on Ubuntu ARM64
- [x] Verify RPM install command on RHEL/Rocky x86_64
- [x] Verify RPM install command on RHEL/Rocky ARM64

### DIFF
--- a/docs/install-deb.md
+++ b/docs/install-deb.md
@@ -24,26 +24,14 @@ References:
 - Fabric Manager: <https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#installing-fabric-manager>
 - NVAT (`nvattest`/`corelib`): <https://docs.nvidia.com/attestation/nv-attestation-sdk-cpp/latest/overview.html>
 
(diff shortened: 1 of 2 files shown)


ebalduf:
What if the user is not downloading this on the machine they are installing on?  A typical scenario is that and administrator of a large fleet with use their laptop (macOS) to download the install package and push it to wherever they install from and then install once from that location on a hundreds of machines!

PR #153, opened 2026-04-01
fix: isolate exporter timeout phases

jingxiang-z:
## Summary
Fix exporter timeout handling so a stalled collection phase does not force HTTP upload to inherit an expired context.

## What Changed
- split exporter collection and upload into separate timeout-scoped contexts
- stop collector work promptly when cancellation or deadline expiry is observed
- preserve partial event data collected before cancellation
- add regression tests for partial collection and fresh upload context behavior

## Root Cause
The exporter previously used one shared timeout for metadata refresh, collection, and upload. When GPU-facing collection work stalled, the shared context expired and the later event-read and HTTP-upload phases inherited a dead context, causing cascading `context deadline exceeded` failures.

## Impact
The agent now degrades more cleanly under driver/NVML stalls: partial collection can still be uploaded, and cancellation no longer produces misleading per-phase timeout cascades.

## Validation
- `go test ./internal/exporter/...`

### DIFF
--- a/internal/exporter/collector/collector.go
+++ b/internal/exporter/collector/collector.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


jingxiang-z:
Superseded by #154, which uses the repo-conventional branch name fix/exporter-timeout-phases.

PR #157, opened 2026-04-06
fix: harden startup and export under slow DCGM and machine info

jingxiang-z:
## Summary
- prewarm machine info and wait up to 5 seconds on the first export before falling back to cached or empty machine info
- move machine info refresh behind a single-flight cache so later exports never block on slow NVML/DCGM-backed machine info collection
- separate collection and HTTP export timeout contexts so slow collection does not consume the send budget
- bound DCGM initialization during server startup to 1 minute and fall back to a no-op DCGM instance when startup enumeration is too slow

## Behavior Notes
- first export now makes a best effort to include machine info without waiting indefinitely
- later exports reuse cached machine info and refresh it in the background only when stale
- the initial machine-info wait is a one-time gate and now honors collection context cancellation
- one export cycle can now take up to roughly collection timeout plus export timeout because those budgets are separated intentionally
- if DCGM startup takes more than 1 minute, the server continues startup with a no-op DCGM instance and skips DCGM-backed checks instead of hanging

## Testing
- `go test ./...`
- `go test -race ./internal/exporter/collector`
- `go test github.com/NVIDIA/fleet-intelligence-sdk/pkg/nvidia-query/dcgm ./internal/server ./internal/exporter`

Related ticket: GPUHEALTH-1735

### DIFF
--- a/internal/exporter/collector/collector.go
+++ b/internal/exporter/collector/collector.go
@@ -185,7 +185,7 @@ func (c *collector) collectMachineInfo(ctx context.Context, data *HealthData) {
 	}
 
 	if _, ok := c.machineInfoProvider.Get(); !ok {
-		c.machineInfoProvider.WaitForInitialRefresh(initialMachineInfoWait)
+		c.machineInfoProvider.WaitForInitialRefresh(ctx, initialMachineInfoWait)
 	}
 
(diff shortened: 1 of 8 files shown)


jingxiang-z:
@coderabbitai review

mukilsh (approved):
LGTM

PR #163, opened 2026-04-13
security: harden agent against local and network attack vectors

rsampaio:
## Description

Switch the default listener from TCP 127.0.0.1:15133 to a Unix socket (/run/fleetint/fleetint.sock) with 0600 permissions so that access is governed by filesystem ownership rather than firewall rules. TCP remains available via --listen-address for backward compatibility. Client commands (status, inject) transparently support both transports through the new endpoint.NewAgentHTTPClient / AgentBaseURL helpers.

Additional hardening in this commit:
- Validate attestation nonce character set and length before passing it to the nvattest exec call (defense-in-depth against a compromised backend crafting malicious arguments).
- Add --token-file flag to `enroll` so the SAK token can be read from a file or stdin instead of appearing in /proc/<pid>/cmdline.
- Reject --path values that point into restricted system directories (/etc, /usr, /sys, ...) when running in offline mode.
- Restrict KMSG_FILE_PATH to paths under /dev/ and validate DCGM_URL character set to prevent environment-variable-based file read or connection redirection by a local attacker.
- Add sync.WaitGroup to attestation Manager so Stop() blocks until the background goroutine exits, eliminating a data race on the global defaultStateFileFn in tests.
## Checklist

## Summary by CodeRabbit

* **New Features**
  * Accept enrollment token from file or stdin (mutually exclusive with token flag)
  * Default server listens on a Unix-domain socket with socket-based client connectivity

* **Improvements**
  * Offline-mode path validation to forbid restricted system directories and symlinks
  * Stronger nonce validation and safer attestation request handling
  * Unified agent request/health handling and more robust server socket lifecycle

* **Chores**
  * Removed legacy process-runner API and its tests

* **Tests**
  * Added unit tests covering token resolution, unix-socket URL/clients, nonce validation, and server/socket behaviors

* **Documentation**
  * Docs and security guidance updated to reflect socket-first defaults and token-file usage


cmd/fleetint/compact.go
```diff
@@ -52,9 +54,18 @@ func compactCommand(cliContext *cli.Context) error {
 		}
 	}
 
-	portOpen := netutil.IsPortOpen(config.DefaultHealthPort) // fleetint uses port 15133
-	if portOpen {
-		return fmt.Errorf("fleetint is running on port %d (must be stopped before running compact)", config.DefaultHealthPort)
+	// Check if the agent is reachable via its default transport (unix socket or TCP).
+	if strings.HasPrefix(config.DefaultListenAddress, "/") {
```
jingxiang-z: If it is a uds, should the customer specify the prefix unix:///?

cmd/fleetint/enroll.go
```diff
@@ -35,11 +38,44 @@ var (
 	storeEnrollmentConfig = storeConfigInMetadata
 )
 
+// resolveToken returns the SAK token from --token, --token-file, or stdin.
```
jingxiang-z, line 41: Does this stdin mean user will be prompted for a token?

rsampaio: No this means the standard input for the CLI, for example:

```
echo TOKEN | fleetint --token-file -
```

cmd/fleetint/compact.go
rsampaio: it is not necessary, we are checking for a specific format to listen on UDS vs TCP and in the same way that the customer only set `127.0.0.1:port` instead of `tcp://host:port` for listening on a socket

jingxiang-z: So the customer will provide /run/fleetint/fleetint.sock as input?

cmd/fleetint/enroll.go
jingxiang-z: Wondering is the reason for this to avoid the token in the command?

rsampaio: The token shows up in the process list with `top` or `ps`, it is not a big deal but it is treated as a secret so it can be leaked

cmd/fleetint/compact.go
rsampaio: In this section it would be `--listen-address /path` but I just caught an inconsistency, the server was expecting only path `/var/run/fleetint/fleetint.sock` but the client was expecting with the schema when we called `--server-url` so I just pushed a fix consolidating this so now both are accepted! Good catch!

jingxiang-z (approved):

PR #169, opened 2026-04-15
security: disable HTTP redirect following on attestation, enrollment, and CLI clients [1/5]

rsampaio:
The exporter client already has CheckRedirect hardened (#165, #168). This covers the remaining three outbound http.Client instances that still follow redirects by default:

- attestation nonce endpoint (also adds status code validation and error-field handling so a non-200 or error response is not silently accepted as a valid nonce)
- enrollment endpoint
- local CLI agent client (NewAgentHTTPClient, used by status/inject)

A compromised backend returning a 302 redirect could bounce these clients into making authenticated requests to internal services. Setting CheckRedirect to return http.ErrUseLastResponse makes the client return the redirect response as-is instead of following it.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP clients in attestation, endpoint, and enrollment now return redirect responses instead of following them.
  * Non-200 HTTP statuses and server-side error payloads are now treated and reported as errors.

* **Tests**
  * Added tests verifying redirects are not followed and that error responses and non-OK statuses are handled as errors.

### DIFF
--- a/internal/attestation/attestation.go
+++ b/internal/attestation/attestation.go
@@ -322,8 +322,14 @@ func (m *Manager) getNonce(jwtToken string, machineId string) (string, time.Time
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", jwtToken))
 
-	// Make the HTTP request
-	client := &http.Client{Timeout: 30 * time.Second}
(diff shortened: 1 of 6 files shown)


jingxiang-z (approved):
Before merge, would like to see if go test -race passed.

PR #170, opened 2026-04-15
security: use absolute paths for exec to prevent PATH poisoning [2/5]

rsampaio:
LocateExecutable relied on exec.LookPath which could return a relative path if "." is in PATH. When the agent runs as root, a local attacker who places a malicious binary in the working directory could hijack execution.

Harden LocateExecutable to resolve the result to an absolute path via filepath.Abs before returning it. Also resolve "sudo" through LocateExecutable in GetSystemManufacturer instead of using a bare command name, so the same absolute-path guarantee applies.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fail early with clearer errors when required system binaries cannot be located.
  * Reject non-absolute (relative) resolved executable paths to avoid running unintended commands.
  * Use resolved absolute executable paths for system and privilege-helper invocations rather than relying on ambiguous fallbacks.

* **Tests**
  * Added a unit test validating rejection of relative executable path results.
  * Broadened CI test tolerance for additional "not found" messages.

### DIFF
--- a/third_party/fleet-intelligence-sdk/pkg/host/virtualization_environment.go
+++ b/third_party/fleet-intelligence-sdk/pkg/host/virtualization_environment.go
@@ -97,7 +97,7 @@ func GetSystemManufacturer(ctx context.Context) (string, error) {
 		return "", nil
 	}
 
-	output, err := exec.CommandContext(ctx, "sudo", dmidecodePath, "-s", "system-manufacturer").CombinedOutput()
+	output, err := exec.CommandContext(ctx, dmidecodePath, "-s", "system-manufacturer").CombinedOutput()
 	if err != nil {
 		return "", fmt.Errorf("failed to read dmidecode for system manufacturer: %w\n\noutput:\n%s", err, strings.TrimSpace(string(output)))
 	}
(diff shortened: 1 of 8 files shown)


third_party/fleet-intelligence-sdk/pkg/host/virtualization_environment.go
```diff
@@ -97,7 +97,11 @@ func GetSystemManufacturer(ctx context.Context) (string, error) {
 		return "", nil
 	}
 
-	output, err := exec.CommandContext(ctx, "sudo", dmidecodePath, "-s", "system-manufacturer").CombinedOutput()
+	sudoPath, err := file.LocateExecutable("sudo")
+	if err != nil {
+		return "", fmt.Errorf("sudo not found (required to run dmidecode): %w", err)
+	}
```
jingxiang-z: I'm thinking we can remove sudo from this command because agent is running as root already. If exec returns non 0 code, just return an error

rsampaio: Agree! I pushed a version without sudo!

jingxiang-z (approved):

PR #173, opened 2026-04-15
security: harden output file creation with symlink checks and 0600 mode [5/5]

rsampaio:
When the agent runs as root, output files were created via os.Create which uses the process umask (typically 0022), producing world-readable files. Additionally, a local attacker could place a symlink at an expected output path to redirect writes to a sensitive system file.

Add safeCreateFile() in both the file writer and CSV converter that:
- Checks for symlinks via os.Lstat before opening (rejects symlinks)
- Creates files with explicit 0600 mode (owner read/write only)

Also chmod the audit log file to 0600 after lumberjack creates it.

## Checklist

## Bug Fixes

* **Enhanced file security**: Output files (CSV exports, JSON exports, and audit logs) are now created with restricted permissions (0600) to prevent unauthorized access.
* **Symlink protection**: Added safeguards to prevent writing through symbolic links, protecting against symlink-based attacks on exported data and logs.

### DIFF
--- a/third_party/fleet-intelligence-sdk/pkg/log/audit.go
+++ b/third_party/fleet-intelligence-sdk/pkg/log/audit.go
@@ -4,10 +4,13 @@
 package log
 
 import (
+	"errors"
+	"fmt"
 	"os"
 	"strings"
 
 	"github.com/google/uuid"
+	"golang.org/x/sys/unix"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 	"gopkg.in/natefinch/lumberjack.v2"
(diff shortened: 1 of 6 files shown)


third_party/fleet-intelligence-sdk/pkg/log/audit.go
```diff
@@ -4,10 +4,13 @@
 package log
 
 import (
```
jingxiang-z, line 6: Are we using this sdk pkg/log? Would like to see if we can remove this.

rsampaio: Yes we do extensively @jingxiang-z, it may be worth to refactor our logging strategy and remove this from the SDK, the SDK should mostly have the metrics collection packages

PR #176, opened 2026-04-15
feat: add backend client and inventory attestation call

jingxiang-z:
## Summary

Implements the agent-side client and runtime loops for the new backend-facing inventory and attestation architecture described in the design doc.

This PR:
- Adds `internal/backendclient` — typed HTTP client for all five agent endpoints (`/v1/agent/enroll`, `/v1/agent/nodes/{nodeUUID}`, `/v1/agent/nodes/{nodeUUID}/nonce`, `/v1/agent/nodes/{nodeUUID}/attestation`, `/v1/agent/token`)
- Adds `internal/agentstate` — centralized persistent state access (JWT, SAK, backend base URL, node ID) backed by the existing SQLite metadata database, with legacy endpoint fallback
- Refactors `internal/attestation` — replaces the monolithic `attestation.go` with split `manager`, `collector`, `backend`, `nonce`, and `types` packages using `backendclient` and `agentstate`
- Adds `internal/inventory` — source, sink, and manager for periodic node inventory upsert via `PUT /v1/agent/nodes/{nodeUUID}`
- Wires enrollment to use `backendclient.Enroll` and performs a best-effort inventory sync immediately after enrollment
- Wires both inventory and attestation loops into the server startup path (`startInventoryLoop`, `startAttestationLoop`)
- Removes machine info fields from OTLP resource attributes — inventory fields now go exclusively through `PUT /v1/agent/nodes/{nodeUUID}`
- Moves `AttestationConfig` out of `HealthExporterConfig` to top-level `Config` and adds `InventoryConfig` with `Enabled`/`Interval`/`InitialInterval` controls and env var overrides
- Deletes the legacy `internal/attestation/attestation.go` monolith and its associated tests
- Bumps Go toolchain to 1.26.2

## Validation

```
go test -race ./internal/backendclient/... ./internal/agentstate/... ./internal/attestation/... ./internal/inventory/... ./internal/enrollment/... ./internal/config/... ./cmd/fleetint/...
```

## Follow-up

- Attach `X-GPUHealth-Agent-Mode: direct-inventory-v1` header on OTLP metrics/logs requests to suppress legacy node extraction on the backend
- Add `serial`, `vendor`, `model` fields to `DiskInfo.BlockDevice` per the inventory field classification in the design doc

## Summary by CodeRabbit

* **New Features**
  * Periodic local inventory collection with change-only backend sync and initial post-enroll sync.
  * Modular attestation flow with nonce/evidence providers, CLI evidence collector, and backend submitter.
  * Local agent state persistence for backend URL, JWT, SAK, and node ID.

* **Configuration**
  * New loop controls for inventory and attestation (enable, intervals, initial-interval, timeouts); defaults provided.

* **Other**
  * Go toolchain bumped to 1.26.2; UI/CSV label changed to "Agent Version"; backend endpoint validation allows localhost/loopback HTTP.


internal/enrollment/enrollment.go
```diff
@@ -13,123 +13,141 @@
...
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", sakToken))
+type machineInfoCollectorFunc func(context.Context) (*machineinfo.MachineInfo, error)
+
+func (f machineInfoCollectorFunc) Collect(ctx context.Context) (*machineinfo.MachineInfo, error) {
+	return f(ctx)
+}
 
-	// Make the request
-	resp, err := client.Do(req)
+func syncInventoryOnce(ctx context.Context) error {
```
ambermingxin, line 146: so enroll passes `context.Background()` to workflow, then the workflow runs NVML + backend sync inline with no timeout. what if NVML init or backend request hangs? the cli might hang too.

internal/enrollment/enrollment.go
```diff
@@ -13,123 +13,141 @@
...
+	defer dbRW.Close()
+
+	if err := config.SecureStateFilePermissions(stateFile); err != nil {
+		return fmt.Errorf("failed to secure state database permissions: %w", err)
+	}
+	if err := pkgmetadata.CreateTableMetadata(ctx, dbRW); err != nil {
+		return fmt.Errorf("failed to create metadata table: %w", err)
+	}
+
+	if err := pkgmetadata.SetMetadata(ctx, dbRW, agentstate.MetadataKeySAKToken, sakToken); err != nil {
```
ambermingxin, line 128: enrollment stores SAK and JWT and base URL, but no node uuid, while inventory export requires node uuid; if absent, it returns `ErrNotReady`. the manager will treat`ErrNotReady` as success, so enroll appears successful but initial sync is skipped silently. is this intentional?

internal/agentstate/sqlite.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (s *sqliteState) GetNodeUUID(ctx context.Context) (string, bool, error) {
+	return s.getMetadata(ctx, pkgmetadata.MetadataKeyMachineID)
+}
+
+func (s *sqliteState) SetNodeUUID(ctx context.Context, value string) error {
+	return s.setMetadata(ctx, pkgmetadata.MetadataKeyMachineID, value)
+}
+
+func (s *sqliteState) getMetadata(ctx context.Context, key string) (string, bool, error) {
```
ambermingxin, line 114: nit: each `Get*` in opens and closes a DB handle. both attestation and inventory paths call multiple getters per cycle, adding some (maybe neglect though) overhead. Have you tested the performance for the new agent?

internal/exporter/collector/collector.go
```diff
@@ -92,38 +84,18 @@ func New(
 	eventStore eventstore.Store,
 	componentsRegistry components.Registry,
 	nvmlInstance nvidianvml.Instance,
-	attestationManager *attestation.Manager,
+	_ any,
 	machineID string,
```
ambermingxin, line 82: some args are no longer used but not removed? are these for backward compatibility concern?

ambermingxin (commented):
As this is a big MR and a lot of refactoring, I recommend a throughout agent test in all installations we support, including different attestation scenarios.

internal/enrollment/enrollment.go
jingxiang-z: nodeUUID is written by fleetint run, but not enroll flow.

internal/enrollment/enrollment.go
jingxiang-z: Added timeout 1m to initial inventory sync and inventory loop and attestation loop.

internal/exporter/collector/collector.go
jingxiang-z: Clean these up.

internal/agentstate/sqlite.go
jingxiang-z: This is negligible at current cadence, inventory=1h, attestation=24h, bootstrap retry 5m.

mukilsh (approved):
LGTM given we've tested

PR #194, opened 2026-05-13
feat: create nvml component

ambermingxin:
## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added NVIDIA NVML GPU health monitoring component to the platform.
  * Improved GPU attribute collection with best-effort approach to handle partial failures gracefully.

* **Tests**
  * Added test coverage for NVML component health checks and partial NVML failure scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/fleet-intelligence-agent/pull/194)

### DIFF
--- a/third_party/fleet-intelligence-sdk/components/accelerator/nvidia/nvml/component.go
+++ b/third_party/fleet-intelligence-sdk/components/accelerator/nvidia/nvml/component.go
@@ -1,5 +1,17 @@
-// Copyright 2024 Lepton AI Inc
-// Source: https://github.com/leptonai/gpud
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 // Package nvml reports NVML collection errors as a component health state.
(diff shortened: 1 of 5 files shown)


third_party/fleet-intelligence-sdk/components/accelerator/nvidia/nvml/component.go
```diff
@@ -0,0 +1,265 @@
...
+	c.lastMu.RLock()
+	last := c.lastCheckResult
+	c.lastMu.RUnlock()
+	if last != nil {
+		return last
+	}
+	return c.checkWithErrors(nil)
+}
+
+func (c *component) LastHealthStates() apiv1.HealthStates {
```
mukilsh, line 109: Is it fair to say LastHealthStates can panic before the first machine-info collection? The component is registered/enabled by default, Start is a no-op, and lastCheckResult starts nil. Health export and the HTTP health handler both call LastHealthStates directly, so if they run before GetMachineGPUInfo calls RunCheckWithErrors, this dereferences nil:

    return last.HealthStates()

Can we make this mirror Check() and return c.checkWithErrors(nil).HealthStates(), or return an explicit healthy/no-data state when last == nil?

mukilsh (commented):
Just left one comment!

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/nvml/component.go
ambermingxin: before first machine-info collection, `LastHealthStates()` returns `healthy / "no data yet"`, not panic.

I would not recommend `c.checkWithErrors(nil).HealthStates()` for this path, because that turns a read API into a mutating check path (updates cached result/timestamp and can add side effects). 

also we cannot implement the Check() function to accept a parameter `collectionErrors`, because `Check()` is defined by the shared `components.Component` interface, so every component must match this exact signature:
`Check() components.CheckResult`
If we change NVML to `Check(collectionErrors []string)`, it no longer implements components.Component, and registration/export paths break.

mukilsh: Okay, this is fine. Looks good then

mukilsh (approved):
LGTM

PR #198, opened 2026-05-19
fix: preserve counter metric types in OTLP

jingxiang-z:
## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added explicit metric-type tracking (counter vs gauge) and a settable counter metric collector.

* **Improvements**
  * OTLP export now emits counters as cumulative monotonic sums and preserves gauges.
  * Scraper and storage now persist and populate metric types; several hardware/network metrics reclassified from gauge → counter.
  * Removed default unit value from the agent "up" metric.

* **Tests**
  * Added tests for type propagation, OTLP conversion, scraper, collector, and schema migration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/fleet-intelligence-agent/pull/198?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/third_party/fleet-intelligence-sdk/pkg/metrics/store/sqlite.go
+++ b/third_party/fleet-intelligence-sdk/pkg/metrics/store/sqlite.go
@@ -43,6 +43,9 @@ const (
 
 	// columnMetricValue represents the numeric value of the metric.
 	columnMetricValue = "metric_value"
+
+	// columnMetricType represents the Prometheus metric type.
+	columnMetricType = "metric_type"
 )
 
 // TODO: drop the old table "gpud_metrics"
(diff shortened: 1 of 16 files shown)


third_party/fleet-intelligence-sdk/pkg/metrics/store/sqlite.go
```diff
@@ -240,9 +256,12 @@ FROM %s
 	for queryRows.Next() {
 		m := pkgmetrics.Metric{}
 		var labels sql.NullString
-		if err := queryRows.Scan(&m.UnixMilliseconds, &m.Component, &m.Name, &labels, &m.Value); err != nil {
+		if err := queryRows.Scan(&m.UnixMilliseconds, &m.Component, &m.Name, &labels, &m.Value, &m.Type); err != nil {
 			return nil, err
 		}
+		if m.Type == "" {
```
rsampaio, line 262: Does this condition cover the upgrade of the agent scenario? I am assuming the customer running a previous version of the agent and upgrade to this version will have this field as "" for existent stored metrics, this will assume the metric is stored as Gauge (which is the type we have been storing).

This seems to handle that case and I want to make sure we are covering the upgrade case.

rsampaio (approved):
Looks good and my only concern is the migration path, if we can test between versions I think this is good to merge

third_party/fleet-intelligence-sdk/pkg/metrics/store/sqlite.go
jingxiang-z: Yes, this covers the upgrade case. For existing stored metrics from older agent versions, metric_type is either added with default gauge during store initialization, or any empty Type is normalized to MetricTypeGauge when reading/writing. In the OTLP converter, only counter is handled specially; empty or non-counter types fall through to Gauge, which matches the previous behavior.

PR #199, opened 2026-05-20
fix: OTLP export silently skips when endpoints are configured but no auth token is present.

miconeilaws:
## Description
Bug: OTLP export silently skips when endpoints are configured but no auth token is present.
  
The README lists OTLP remote export as a supported format and the metadata store allows configuring metrics_endpoint/logs_endpoint independently. However, exportToHTTP has an early return when AuthToken == "" that silently drops all data even though the HTTP writer already handles missing tokens correctly (just omits the header).

This removes the redundant gate so the export path works as documented when endpoints are configured.

## Checklist
- [X] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/fleet-intelligence-agent/blob/main/CONTRIBUTING.md).
- [X] New or existing tests cover these changes.
- [X] The documentation is up to date with these changes.

## Summary by CodeRabbit

* **New Features**
  * HTTP export proceeds when no authentication token is configured; a warning is logged if endpoints are present but the token is empty.
  * If the server returns an updated token, the client adopts and attempts to persist it; export still completes even if persistence fails.

* **Tests**
  * Added tests validating export without an auth token and verifying token-refresh and persistence behavior via a test server.

### DIFF
--- a/internal/exporter/exporter.go
+++ b/internal/exporter/exporter.go
@@ -243,8 +243,7 @@ func (e *healthExporter) exportToHTTP(ctx context.Context, data *collector.Healt
 	}
 
 	if e.options.config.AuthToken == "" {
-		log.Logger.Infow("No auth token configured, skipping HTTP export")
-		return nil
+		log.Logger.Warnw("No auth token configured, exporting without authentication")
 	}
 
 	newToken, err := e.httpWriter.Send(ctx, data, e.options.config.MetricsEndpoint, e.options.config.LogsEndpoint, e.options.config.RetryMaxAttempts, e.options.config.AuthToken)
--- a/internal/exporter/exporter_test.go
+++ b/internal/exporter/exporter_test.go
@@ -25,6 +25,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


jingxiang-z:
@miconeilaws The change looks good to me. Could you please rebase this branch on the latest main?
Also, could you share your use case? Are you trying to connect the agent to a local OTEL Collector, or to a remote OTLP endpoint that does not require auth?

internal/exporter/exporter.go
```diff
@@ -189,26 +176,6 @@ func (e *healthExporter) export() error {
...
-
-	computeZone, err := pkgmetadata.ReadMetadata(ctx, e.options.dbRO, agentstate.MetadataKeyComputeZone)
-	if err != nil {
-		log.Logger.Debugw("compute zone metadata not available for telemetry resource", "error", err)
-	} else {
-		data.ComputeZone = computeZone
-	}
-}
-
 // exportToFile writes health data to files
```
rsampaio, line 212: Thanks again for this PR! 

I believe it would be valuable to keep a warning message about the lack of token but remove the return statement so it is informative and will keep the purpose of this change. 

We intend to release an OTEL gateway that would allow to capture metrics and feed a custom dashboard/metrics system but this change is still a good catch!

miconeilaws:
Thanks guys, rebased and kept the warning log.

I'm trying to forward the telemetry to my existing OTel node-local collector so I can ingest it in my existing telemetry pipeline. 

Ideally I would also want a TLS config for auth as thats what I do for other things like scraping prometheus endpoints. I did evaluate the prometheus option here but it doesn't give me the health classifications as those seem to get emitted as logs. 

No auth is fine for some prototyping for now.

miconeilaws:
@jingxiang-z @rsampaio  let me know of any other blockers

jingxiang-z:
@miconeilaws Thanks. Could you rebase the latest main to include the security fix? Also could you change the PR title to be fix: ...

miconeilaws:
@jingxiang-z done, thanks a lot!

jingxiang-z (approved):

jingxiang-z:
@miconeilaws Thanks for your contribution! Approved and merged.

PR #203, opened 2026-05-22
fix: context deadline exceeded shouldn't throw an error

ambermingxin:
## Checklist

## Summary by CodeRabbit

* **New Features**
  * Offline mode now performs an immediate export at startup so short-lived runs produce output before the first scheduled interval.

* **Improvements**
  * Server shutdown now coordinates and waits for background loops to exit, with bounded wait and clearer timeout handling.

* **Documentation**
  * Helm chart README, values, daemonset template, and install guide updated with optional enrollment metadata flags and usage notes.

* **Tests**
  * Added tests for offline export behavior and for background-loop coordination and shutdown.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/fleet-intelligence-agent/pull/203?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
+++ b/deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
@@ -74,6 +74,12 @@ spec:
               {{- if .Values.enroll.force }}
               --force
               {{- end }}
+              {{- if ne .Values.enroll.nodeGroup nil }}
+              --node-group {{ .Values.enroll.nodeGroup | quote }}
+              {{- end }}
+              {{- if ne .Values.enroll.computeZone nil }}
+              --compute-zone {{ .Values.enroll.computeZone | quote }}
+              {{- end }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           env:
--- a/deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
+++ b/deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
@@ -74,10 +74,10 @@ spec:
               {{- if .Values.enroll.force }}
               --force
               {{- end }}
-              {{- if ne .Values.enroll.nodeGroup nil }}
+              {{- if hasKey .Values.enroll "nodeGroup" }}
               --node-group {{ .Values.enroll.nodeGroup | quote }}
               {{- end }}
-              {{- if ne .Values.enroll.computeZone nil }}
+              {{- if hasKey .Values.enroll "computeZone" }}
               --compute-zone {{ .Values.enroll.computeZone | quote }}
               {{- end }}
           securityContext:
--- a/deployments/helm/fleet-intelligence-agent/values.yaml
+++ b/deployments/helm/fleet-intelligence-agent/values.yaml
@@ -61,10 +61,13 @@ enroll:
   tokenSecretName: ""
   tokenSecretKey: "token"
   tokenValue: ""
-  # Optional enrollment metadata. Keep null to omit flag (preserve existing value).
-  # Set to a string to overwrite. Set to empty string "" to clear.
-  nodeGroup: ""
-  computeZone: ""
+  # Optional enrollment metadata:
+  # - omit key to preserve existing stored value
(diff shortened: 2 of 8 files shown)


internal/exporter/exporter.go
```diff
@@ -116,6 +116,16 @@ func (e *healthExporter) Start() error {
 
 	log.Logger.Infow("Starting health exporter")
 
+	// In offline mode, emit one export immediately so short runs don't exit
```
jingxiang-z, line 119: Need to check do we have all data available if we export immediately

internal/server/server.go
```diff
@@ -210,6 +210,16 @@ func getAttestationTimeout(cfg *config.Config) time.Duration {
 	return config.DefaultAttestationTimeout
 }
 
+func shouldLogLoopExitError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
jingxiang-z: What if there is a real deadline exceed error, in this case, we won't throw error log? Is there a way we can differentiate these two scenarios?

deployments/helm/fleet-intelligence-agent/values.yaml
```diff
@@ -61,6 +61,10 @@ enroll:
   tokenSecretName: ""
   tokenSecretKey: "token"
   tokenValue: ""
+  # Optional enrollment metadata. Keep null to omit flag (preserve existing value).
+  # Set to a string to overwrite. Set to empty string "" to clear.
+  nodeGroup: ""
+  computeZone: ""
```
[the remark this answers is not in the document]
ambermingxin: it is an example values.yaml that default these values as empty.

internal/server/server.go
```diff
@@ -210,6 +216,34 @@ func getAttestationTimeout(cfg *config.Config) time.Duration {
...
 
+func shouldLogLoopExitError(ctx context.Context, err error) bool {
+	if err == nil {
+		return false
+	}
+	if ctx != nil && ctx.Err() != nil && (errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)) {
+		return false
+	}
+	return true
+}
```
jingxiang-z: If we have waitGroup, do we still need this?

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
```diff
@@ -74,6 +74,12 @@ spec:
               {{- if .Values.enroll.force }}
               --force
               {{- end }}
+              {{- if ne .Values.enroll.nodeGroup nil }}
+              --node-group {{ .Values.enroll.nodeGroup | quote }}
+              {{- end }}
+              {{- if ne .Values.enroll.computeZone nil }}
+              --compute-zone {{ .Values.enroll.computeZone | quote }}
+              {{- end }}
```
jingxiang-z: Can we add guardrail on both these need to be set on the helm install install level

internal/server/server.go
ambermingxin: I think it's no harm to keep this. `WaitGroup` resolves ordering like don’t tear down before goroutines finish to make sure the lifecycle correctness. `shouldLogLoopExitError` solves classification (“is this exit expected or a real failure?”) and can be good for log signal. 

Even with graceful shutdown: loops can return context.Canceled / context.DeadlineExceeded on normal stop signal. without filter, we'd still emit error logs during expected shutdown.

jingxiang-z: I'm worried about this hides real context deadline exceeded error. We had the issue that the driver hangs cause the inventory loop to fail. If we hide this error message, customer and we have no way to debug this kind of issue.

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
ambermingxin: these fields are not required. i improved the guard to use `hasKey`

jingxiang-z (approved):

PR #218, opened 2026-06-11
fix: ubuntu deb package upgrade restart the service

jingxiang-z:
## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined install/upgrade behavior for the bundled service: fresh installs enable and start the service, while upgrades and configuration runs now only restart the service if it is already running, avoiding unintended service starts. This reduces unexpected service restarts and improves stability during package operations.

### DIFF
--- a/deployments/packages/scripts/postinst
+++ b/deployments/packages/scripts/postinst
@@ -21,11 +21,11 @@ else
         configure)
             if [ -n "$2" ]; then
                 # $2 is the previous version — Debian upgrade path
-                echo "Upgrade detected - (re)starting service..."
+                echo "Upgrade detected - restarting service if already running..."
                 if systemctl is-active --quiet fleetintd.service; then
                     systemctl restart fleetintd.service || echo "[WARNING] Failed to restart service"
                 else
-                    systemctl start fleetintd.service || echo "[WARNING] Failed to start service"
+                    echo "Service is not running - not starting during upgrade"
                 fi
             else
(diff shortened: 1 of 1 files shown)


deployments/packages/scripts/postinst
```diff
@@ -18,17 +18,32 @@ else
     systemctl daemon-reload || echo "[WARNING] Failed to reload systemd daemon"
 
     case "$1" in
-        configure|install|1)
+        configure)
+            if [ -n "$2" ]; then
+                # $2 is the previous version — Debian upgrade path
+                echo "Upgrade detected - restarting service if already running..."
+                if systemctl is-active --quiet fleetintd.service; then
+                    systemctl restart fleetintd.service || echo "[WARNING] Failed to restart service"
```
rsampaio, line 26: @jingxiang-z we should consider using [deb-systemd-invoke](https://manpages.debian.org/testing/init-system-helpers/deb-systemd-invoke.1p.en.html) for the postinst script, it is recommended for package scripts since it respects debian polices and prevent invocation of systemctl in containers since there is not init running

jingxiang-z: Thanks for the suggestion! deb-systemd-invoke is useful but since the postinst script is shared by both deb package and rpm package, there is no clean split unless we have condition on the package type. I can look into separate the pre/post script for rpm and deb though.

rsampaio (approved):

PR #220, opened 2026-06-23
fix: make DCGM group name configurable and different for each process

jingxiang-z:
## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Updated GPU/DCGM monitoring to use explicit, process-/owner-derived DCGM group and field-group names instead of defaults or hard-coded identifiers, improving isolation across concurrent processes.
  * Aligned GPU field-watching with the selected named groups, and ensured reconnect behavior reuses the same group naming.
  * Standardized default GPU field-group naming for field caching.
* **Tests**
  * Added a unit test covering cleanup when DCGM group creation fails.

### DIFF
--- a/internal/scan/scan.go
+++ b/internal/scan/scan.go
@@ -124,8 +124,10 @@ func Scan(ctx context.Context, opts ...Option) error {
 		}
 	}
 
+	dcgmGroupNames := components.NewDCGMGroupNames(fmt.Sprintf("scan-%d", os.Getpid()))
+
 	// Initialize DCGM instance for DCGM-based health checks
-	dcgmInstance, err := nvidiadcgm.New()
+	dcgmInstance, err := nvidiadcgm.NewWithGroupName(dcgmGroupNames.HealthMonitoringGroup)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


ambermingxin (approved):

internal/scan/scan.go
```diff
@@ -124,8 +124,10 @@ func Scan(ctx context.Context, opts ...Option) error {
 		}
 	}
 
+	dcgmGroupNames := components.NewDCGMGroupNames(fmt.Sprintf("scan-%d", os.Getpid()))
+
 	// Initialize DCGM instance for DCGM-based health checks
-	dcgmInstance, err := nvidiadcgm.New()
+	dcgmInstance, err := nvidiadcgm.NewWithGroupName(dcgmGroupNames.HealthMonitoringGroup)
```
rsampaio, line 130: do we clean up the session with the GroupName at some point? I'm asking because PIDs can be reused and we may end up with the same PID depending on the kernel configuration, random PID is often the default and the risk of reuse is high on long running servers

rsampaio (approved):

internal/scan/scan.go
jingxiang-z: Yes, when each process(scan/daemon) finished, we cleaned up the DCGM group we created.

PR #224, opened 2026-06-25
fix: initialize node uuid during enrollment

jingxiang-z:
## Checklist

## Summary by CodeRabbit

* **New Features**
  * Centralized node UUID “load or create” is now used during enrollment and server startup, persisted in agent state.
  * Node UUID prefers hardware product UUID (DMI) when available, with fallback to a generated UUID for reuse.
* **Bug Fixes**
  * Enrollment now fails fast if node UUID initialization cannot complete, preventing subsequent post-enroll steps.
  * Re-runs are idempotent and can repair empty stored values.
* **Deployment**
  * Helm DaemonSet enroll init step now mounts host `/sys` read-only.
* **Tests**
  * Expanded unit and workflow coverage for UUID persistence, reuse, hardware/DMI reading, and error propagation.

### DIFF
--- a/internal/agentstate/sqlite.go
+++ b/internal/agentstate/sqlite.go
@@ -119,8 +119,7 @@ func (s *sqliteState) GetOrCreateNodeUUID(ctx context.Context, create func() (st
 	}
 	defer db.Close()
 
-	stateFile, err := s.stateFileFn()
-	if err == nil {
+	if stateFile, err := s.stateFileFn(); err == nil {
 		if err := config.SecureStateFilePermissions(stateFile); err != nil {
 			return "", false, fmt.Errorf("secure state file permissions: %w", err)
 		}
@@ -254,8 +253,7 @@ func (s *sqliteState) setMetadata(ctx context.Context, key, value string) error 
 	}
 	defer db.Close()
 
-	stateFile, err := s.stateFileFn()
-	if err == nil {
+	if stateFile, err := s.stateFileFn(); err == nil {
 		if err := config.SecureStateFilePermissions(stateFile); err != nil {
 			return fmt.Errorf("secure state file permissions: %w", err)
 		}
--- a/internal/agentstate/sqlite.go
+++ b/internal/agentstate/sqlite.go
@@ -112,6 +112,86 @@ func (s *sqliteState) SetNodeUUID(ctx context.Context, value string) error {
 	return s.setMetadata(ctx, pkgmetadata.MetadataKeyMachineID, value)
 }
 
+func (s *sqliteState) GetOrCreateNodeUUID(ctx context.Context, create func() (string, error)) (string, bool, error) {
+	db, err := s.openReadWrite()
+	if err != nil {
+		return "", false, err
+	}
+	defer db.Close()
+
+	stateFile, err := s.stateFileFn()
+	if err == nil {
+		if err := config.SecureStateFilePermissions(stateFile); err != nil {
+			return "", false, fmt.Errorf("secure state file permissions: %w", err)
+		}
+	}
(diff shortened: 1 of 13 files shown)


internal/agentstate/sqlite.go
```diff
@@ -112,6 +112,83 @@ func (s *sqliteState) SetNodeUUID(ctx context.Context, value string) error {
...
+		candidate,
+	); err != nil {
+		return "", false, fmt.Errorf("insert node UUID metadata: %w", err)
+	}
+
+	committed, ok, err := readNodeUUIDDB(ctx, db)
+	if err != nil {
+		return "", false, err
+	}
+	if !ok {
```
ambermingxin: it's better to turn it into if-else statement for code readability

internal/agentstate/state.go
```diff
@@ -42,6 +42,7 @@ type State interface {
 
 	GetNodeUUID(ctx context.Context) (value string, ok bool, err error)
 	SetNodeUUID(ctx context.Context, value string) error
+	GetOrCreateNodeUUID(ctx context.Context, create func() (string, error)) (value string, created bool, err error)
```
ambermingxin, line 47: This function seems to handle get situation too. can we remove `GetNodeUUID` if it has duplicate scenario? if not better to add a comment to differentiate them.

internal/server/server.go
```diff
@@ -137,40 +137,80 @@ func initializeDatabases(ctx context.Context, cfg *config.Config) (*sql.DB, *sql
...
+	if !ok {
+		return "", false, fmt.Errorf("node UUID metadata was not committed")
+	}
+	if committed == "" {
+		return "", false, fmt.Errorf("node UUID metadata is empty")
+	}
+	return committed, committed == candidate, nil
+}
+
+func readNodeUUIDDB(ctx context.Context, db *sql.DB) (string, bool, error) {
```
ambermingxin: `GetOrCreateNodeUUID` and `readNodeUUIDDB` are duplicated nearly verbatim in two places. and the identical sql flow in `internal/agentstate/sqlite.go`. a future fix to one is easy to miss the other. I'd extract a shared helper that operates on a `*sql.DB`, e.g. `metadata.GetOrCreateMachineID(ctx, db, create)`, and have both call sites delegate to it. The only nuance (agentstate version also calls `SecureStateFilePermissions`) can stay in the wrapper.

internal/nodeidentity/nodeidentity.go
ambermingxin, line 1: nitpick: code mixes "node UUID" with "machine ID". It's internally consistent enough so it's fine, but a one-line comment noting they refer to the same value would help in future.

internal/agentstate/state.go
jingxiang-z: We can’t remove GetNodeUUID because some callers need a read-only check that reports a missing UUID instead of creating and persisting one as a side effect.

internal/agentstate/sqlite.go
```diff
@@ -112,6 +112,86 @@ func (s *sqliteState) SetNodeUUID(ctx context.Context, value string) error {
...
 }
 
+func (s *sqliteState) GetOrCreateNodeUUID(ctx context.Context, create func() (string, error)) (string, bool, error) {
+	db, err := s.openReadWrite()
+	if err != nil {
+		return "", false, err
+	}
+	defer db.Close()
+
+	stateFile, err := s.stateFileFn()
```
rsampaio: nit: this if could probably be tightened since the stateFile is never used outside of the condition:

```go
if stateFile, err := s.stateFileFn(); err == nil {
```

ambermingxin (approved):

PR #231, opened 2026-07-14
feat: add authenticated OpenTelemetry gateway

rsampaio:
## Description

Adds an optional, production-hardened OpenTelemetry gateway deployment for Fleet Intelligence. Agents route OTLP metrics and logs through a per-cluster custom Collector, which owns backend SAK-to-JWT authentication; direct agent enrollment remains unchanged for inventory and attestation.

The gateway is disabled by default and does not change bare-metal or existing Kubernetes behavior unless `otelGateway.enabled=true`.

### Security and reliability

- Keeps the service SAK in a Kubernetes Secret and an opaque Collector config field.
- Requires HTTPS enrollment, rejects redirects, bounds response bodies, and supports custom CA/TLS settings.
- Serializes concurrent JWT refreshes and retries a replayable request at most once after a 401.
- Runs non-root with a read-only filesystem, dropped capabilities, RuntimeDefault seccomp, and no service-account token.
- Defaults to two replicas with rolling updates, topology spreading, a PDB, probes, resource limits, and ingress NetworkPolicy.
- Publishes multi-architecture images to GHCR and builds against OpenTelemetry Collector v0.156.0.

### Review guide

1. **Agent routing:** `cmd/fleetint/run.go`, `internal/config/`, and `internal/exporter/` establish collector precedence without forwarding node credentials.
2. **Gateway authentication:** `otelcol/auth/sakauth/` contains enrollment, token refresh, TLS, and request authentication.
3. **Deployment:** `deployments/helm/fleet-intelligence-agent/templates/otel-gateway-*` contains availability and security controls.
4. **Build/release:** `otelcol/Dockerfile`, `otelcol/otelcol-builder.yaml`, and GitHub workflows build, test, and publish the collector.
5. **Tests:** `collector_integration_test.go` starts the built Collector and verifies real OTLP metrics/logs forwarding, identity attributes, and backend authentication.

### Operational notes

- Agent-to-gateway OTLP is cluster-local HTTP and is restricted by NetworkPolicy; clusters must use a CNI that enforces NetworkPolicy.
- Production installations should use `otelGateway.enroll.existingSecret` rather than passing a SAK through Helm values.
- The sending queue is in-memory; persistent queueing is not introduced by this change.

## Test plan

- [x] Gateway unit tests with race detector (`90.8%` statement coverage).
- [x] End-to-end built-Collector test for SAK enrollment and OTLP metrics/logs forwarding.
- [x] Core config, backend client, endpoint, and agent-state regression tests.
- [x] Gateway module `go vet` and full `golangci-lint` (`0 issues`).
- [x] Standalone Collector configuration validation.
- [x] Helm lint and rendering with gateway disabled and enabled.
- [ ] Full Linux Docker suite and root lint will run in PR CI; local execution is blocked by Docker socket sandboxing and macOS-only build constraints around DCGM/NVML/procfs.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an optional OpenTelemetry gateway to route metrics and logs via a dedicated collector.
  * Extended Helm charts with gateway deployment, authentication, service exposure, and optional network controls.
  * Added support for a configurable collector endpoint to drive gateway routing.
  * Added multi-architecture releases for the `fleetint-otelcol` image, including automated publishing/tagging.
* **Bug Fixes**
  * Improved token protection by consistently masking sensitive metadata values.
  * Added stricter validation for collector/enrollment endpoints and enforced safer URL handling.
* **Tests**
  * Added unit tests for endpoint/masking behavior and an end-to-end gateway/auth integration test.
* **Chores**
  * Updated CI, release, and build tooling to lint/build/test and publish the new collector image.


deployments/helm/fleet-intelligence-agent/templates/otel-gateway-pdb.yaml
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jingxiang-z, line 1: Wondering why do we need a pdb here?

deployments/helm/fleet-intelligence-agent/templates/otel-gateway-networkpolicy.yaml
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jingxiang-z, line 1: Wondering why do we need a network policy here? Is it because we want to do some security hardening to only allow traffic from our agent?

deployments/otel-collector/config.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+processors:
+  memory_limiter:
+    check_interval: 5s
+    limit_mib: 400
+    spike_limit_mib: 100
+
+  batch:
+    timeout: 10s
+    send_batch_size: 1000
```
jingxiang-z: Can this memory limiter be configured through helm values? What if we change the OTLP pod memory configuration to a large value say 2GB, do we also need to change this memory limiter?

otelcol/auth/sakauth/extension.go
```diff
@@ -0,0 +1,287 @@
...
+		return current, nil
+	}
+
+	jwt, err := e.performEnrollment(ctx)
+	if err != nil {
+		return "", err
+	}
+	e.mu.Lock()
+	e.jwt = jwt
+	e.customerID = extractCustomerID(jwt)
```
jingxiang-z: Do we need the customer id from the backend?

.github/workflows/ci.yml
```diff
@@ -177,6 +177,40 @@ jobs:
           path: fleetint-${{ matrix.goos }}-${{ matrix.goarch }}
           retention-days: 7
 
+  build_check_otelcol:
```
ambermingxin, line 192: `otelcol/auth/sakauth` is a separate Go module (it has its own `go.mod`). In `.github/workflows/ci.yml`, `unit_test` runs make test → `go test ./...` from the root module, which does not descend into nested modules. similar to the lint runs `golangci-lint` at the root and vuln runs `govulncheck ./...` at the root. `build_check_otelcol` only runs the single integration test: run `^TestCollectorGatewayEndToEnd$ ` and only for amd64). so `extension_test.go` and config tests are never executed in CI. Should we add something like following in ci to include that? 

```
- name: Test sakauth module
  working-directory: otelcol/auth/sakauth
  run: go test -race ./...
```

deployments/helm/fleet-intelligence-agent/templates/otel-gateway-configmap.yaml
```diff
@@ -0,0 +1,85 @@
...
+        protocols:
+          http:
+            # No inbound auth: the gateway is a ClusterIP service accessible only
+            # within the cluster. Cluster network isolation is the security boundary.
+            # The gateway uses its own SAK JWT when forwarding to the backend.
+            # Agent identity flows through OTLP resource attributes (machine.id, GPU UUIDs).
+            endpoint: 0.0.0.0:{{ .Values.otelGateway.service.port }}
+
+    processors:
+      memory_limiter:
```
ambermingxin, line 58: from default values.yaml, 
```
memory_limiter:
  limit_percentage: 80
  spike_limit_percentage: 20
```
is this limiter computed from the total node mem or the pod limit? if it's the former, there could possibly be OOM issue if the pod is capped at a smaller mem e.g. `resources.limits.memory: 512Mi` while the node itself is large, say 512G

.github/workflows/ci.yml
rsampaio: Good catch! I added to the Makefile so the ci and local tests will include the otelcol directory!

deployments/helm/fleet-intelligence-agent/templates/otel-gateway-configmap.yaml
rsampaio: Pod limit `memorylimiter` reads the cgroup quota, falling back to node memory only when no limit is set. We default to limits.memory: 512Mi, so 80% is ~410Mi.

The fallback is a real risk with resources: {}, so I added a fail guard plus a note in values.yaml.

deployments/helm/fleet-intelligence-agent/templates/otel-gateway-networkpolicy.yaml
rsampaio: Yes exactly! The OTLP receiver has no auth for incoming connections, so without it any pod in the cluster could push telemetry that we then sign with the service SAK. This limits ingress to our own pods on 4318.

deployments/helm/fleet-intelligence-agent/templates/otel-gateway-pdb.yaml
rsampaio: The gateway is 2 replicas on the metrics/logs path and the sending queue is in-memory, so if a drain evicts both at once we drop whatever is in flight. `minAvailable: 1` keeps one up.

It only renders when the gateway is enabled and there is a config for that, we can default to false if that is less confusing

deployments/otel-collector/config.yaml
rsampaio: Good catch! There were two configs disagreeing. The one you're looking at wasn't read by Helm, so I deleted it,  the ConfigMap is the only one now.

It's configurable via `otelGateway.memoryLimiter.*` and it's a percentage, so it follows the pod limit.

jingxiang-z (approved):

otelcol/auth/sakauth/extension.go
rsampaio: I thought it was required but indeed after checking the backend this seems unused, there is an use case fr the resource attribute which I will investigate further and follow up with a new PR, this PR is already too big

rsampaio: I will remove for now to simplify this PR

PR #236, opened 2026-07-24
chore: add linux package and checksum signing

jingxiang-z:
## Checklist

## Summary by CodeRabbit

- **New Features**
  - Linux DEB and RPM packages are now signed and verified before release.
  - Releases include signed checksum manifests, a public verification key, and verification tools.
  - Releases are published only after all artifacts pass authenticity checks.

- **Documentation**
  - Added optional DEB and RPM package verification instructions, including fingerprint, signature, and checksum checks.
  - Updated release guidance for public-key import and checksum verification.

- **Bug Fixes**
  - Improved checksum refreshing and artifact validation.
  - Existing published releases are preserved rather than rebuilt or overwritten.

- **Chores**
  - Updated the project’s Go version to 1.26.6.


docs/install-deb.md
```diff
@@ -63,6 +65,32 @@ sudo fleetint --version
...
+## Optional: Verify package authenticity
+
+To verify a release before installation, download the package,
+`fleet-intelligence.pub.asc`, `checksums.txt`, `checksums.txt.asc`, and
+`verify-linux-package-signature.sh` from the same release.
+
+Confirm that the public key fingerprint is:
+
+```text
+FE0C 8B74 CA66 357C 13BE 197D CCE3 C963 0871 99B3
```
ambermingxin, line 77: is this key an existing, already-published NVIDIA release key authoritative outside this repo? If so, better to link it here in md.

scripts/refresh-checksums.sh
ambermingxin, line 1: if i understand correctly, the steps are: sign packages -> refresh checksums -> sign the manifest? this script re-hashes every artifact listed in `checksums.txt` and overwrites each entry unconditionally. It never compares against what GoReleaser computed. if i seed a wrong hash for an artifact that the signing step never touches, and the script will silently corrected it and exited 0

docs/install-deb.md
jingxiang-z: The public key will be published in this repo and release artifact.

scripts/refresh-checksums.sh
jingxiang-z: So the goreleaser generated deb/rpm package is not signed, after it is signed, the checksum will change, so the pipeline has to sign the changed checksum.

ambermingxin: I agree with the chain: rewrites the deb/rpm and SHA-256 changes, so the manifest has to be regenerated before it's signed. I wasn't asking whether the refresh step should exist - the step in the workflow sounds about right. My point was about its scope. `sign-linux-packages.sh` only check deb/rpm, so those are the only files whose hashes can legitimately change. The tarballs and every other GoReleaser artifact must be identical. But `refresh-checksums.sh` re-hashes every line unconditionally, so it can't tell "expected change" from "unexpected change"

deployments/packages/keys/fleet-intelligence.pub.asc
rsampaio, line 1: How are we managing this PGP key? We should make sure that when it expires we update the keys and we should probably monitor this as well

docs/install-deb.md
jingxiang-z: This is a Fleet Intelligence-specific signing key, not an existing NVIDIA-wide release key published by another authoritative service. The private-key operation is handled by NVIDIA 3S; this repository contains only the public key. The authoritative public distribution points for this key are the tracked file in this repository and the copy attached to each GitHub release. The documented full fingerprint is the trust anchor users should compare before importing it, so there is no separate external key URL to link today.

scripts/refresh-checksums.sh
jingxiang-z: Agreed—the current refresh scope is too broad. Only the DEB/RPM files are replaced by the signing step, so only those checksum entries should be regenerated. For every other GoReleaser artifact, the script should verify the file against the original checksum and fail if it changed, then add the explicitly supplied public key and verifier entries before signing the final manifest. That preserves GoReleaser's integrity check for untouched artifacts instead of silently accepting unexpected changes.

deployments/packages/keys/fleet-intelligence.pub.asc
jingxiang-z: The private key is managed by NVIDIA 3S and is never stored in this repository. The repository contains only the public key, with its full fingerprint pinned in the verification scripts. The current key has no expiration timestamp, so there is no expiry date to monitor as issued. We should still document ownership and a rotation/revocation procedure: provision the replacement in 3S, publish the new public key and fingerprint, update the scripts and documentation, validate a release end to end, and retain the old public key for historical releases. If a future key has an expiry, we should add advance-expiry monitoring.

.github/workflows/release.yml
rsampaio: Could we pass `github.ref_name` using a named environment variable and validate the quoted shell variable? I realize this interpolation predates the PR, but GitHub expands the expression into the script before Bash parses it, and git permits $(), braces, and $ in tag names and a crafted v* tag can execute command substitution before the regex check. This PR raises the impact because the same self-hosted job now receives the 3S client secret and controls signed artifacts. I would also gate this job with protected release tags.

.github/workflows/release.yml
```diff
@@ -169,41 +263,31 @@ jobs:
...
-        run: |
-          jf config add nvidia \
-            --url="${{ secrets.ARTIFACTORY_URL }}" \
-            --user="${{ secrets.ARTIFACTORY_USER }}" \
-            --access-token="${{ secrets.ARTIFACTORY_TOKEN }}" \
-            --interactive=false \
-            --overwrite
-          jf config use nvidia
-      
-      - name: Upload x86_64 tarball to Artifactory
```
rsampaio, line 189: Will this impact previous uploads to artifactory? Should we remove the existent ones?

.github/workflows/release.yml
jingxiang-z: Thanks, added a env var to validate the tag and also added a tag rule, only admin and maintainer of this repo are able to push tag.

.github/workflows/release.yml
jingxiang-z: We uploaded it before to use internal pipeline to create deb/rpm package. But we have our own deb/rpm creation pipeline and never used the NVIDIA one. So it's safe to deprecate this. Also I received an email that the internal artifactory is deprecated so we shouldn't upload the artifact there any more.

.github/workflows/release.yml
```diff
@@ -39,31 +49,115 @@ jobs:
...
-      
+
+      - name: Verify nvsec is available
+        run: |
+          command -v nvsec >/dev/null 2>&1 || {
+            echo "nvsec must be installed on the self-hosted release runner" >&2
+            exit 1
+          }
+
+      - name: Sign and verify DEB and RPM packages
```
rsampaio, line 92: Let's validate that a rerun wouldn't replace already signed packages, this can give different results for a customer that downloaded a version and this version is replaced.

.github/workflows/release.yml
```diff
@@ -17,6 +18,15 @@ jobs:
...
     steps:
+      - name: Validate tag format
+        shell: bash
+        run: |
+          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9]+(\.[a-zA-Z0-9]+)*)?$ ]]; then
+            printf 'Tag %q does not match expected format (v1.2.3, v1.2.3-beta, or v1.2.3-rc.2)\n' \
+              "$RELEASE_TAG" >&2
+            exit 1
+          fi
+
```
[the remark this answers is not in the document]
jingxiang-z: v* is protected by a tag ruleset. Only admin and maintainer are able to create tag.

.github/workflows/release.yml
jingxiang-z: Thanks, made a change to reject the same tag name and concurrent run.

rsampaio (approved):

PR #249, opened 2026-08-26
feat: prototype shared GPU collection integration

ooolafhu:
## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/dsx-ai-factory/fleet-intelligence-agent/blob/main/CONTRIBUTING.md).


internal/sharedcollect/inventory.go
```diff
@@ -0,0 +1,275 @@
...
+	nvidiaproduct "github.com/NVIDIA/fleet-intelligence-sdk/pkg/nvidia/product"
+
+	"github.com/dsx-ai-factory/health-validation/collect/observation"
+
+	"github.com/NVIDIA/fleet-intelligence-agent/internal/inventory"
+)
+
+// GPUInventoryFromObservations reconstructs FI's existing GPU inventory
+// model. Values are merged by UUID: DCGM owns the metric-compatible GPU index,
+// while NVML is preferred for the remaining inventory fields.
```
jingxiang-z: What inventory fields are only available in NVML?

internal/sharedcollect/collector_linux.go
```diff
@@ -0,0 +1,160 @@
...
+func New(options Options) (*Collector, error) {
+	if options.SourceTimeout <= 0 {
+		options.SourceTimeout = defaultCollectionTimeout
+	}
+	nvmlInstance, err := nvmlsource.New()
+	if err != nil {
+		return nil, fmt.Errorf("initialize shared NVML source: %w", err)
+	}
+
+	dcgmSource := dcgmsource.NewSource(dcgmsource.Options{
```
jingxiang-z: Should this be consistent naming? nvmlSource and dcgmSource?

internal/sharedcollect/collector_unsupported.go
```diff
@@ -0,0 +1,34 @@
+//go:build !linux || !cgo
```
jingxiang-z: I think we only support linux

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/dcgm/inforom/component.go
```diff
@@ -84,11 +84,6 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 		} else {
 			log.Logger.Infow("registered DCGM inforom health watch")
 		}
-		if err := c.dcgmInstance.AddFieldsToWatch(inforomFields); err != nil {
```
jingxiang-z, line 87: Why do we just remove the add fields to watch?
The end goal of this is to remove this third_party package

internal/sharedcollect/inventory.go
ooolafhu: I believe Board ID is missing from DCGM: https://github.com/dsx-ai-factory/fleet-intelligence-agent/blob/96ced8caf1d256352b4d27e13133c25c26aaa2a5/third_party/fleet-intelligence-sdk/pkg/machine-info/machine_info.go#L385-L392

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/dcgm/inforom/component.go
ooolafhu: Yes that's the end goal. Since shared library doesn't have a good coverage of what third_party package is doing yet, I kept most the code unchanged. Just deleted the `AddFieldsToWatch` to confirm that metrics are collected via shared library. 

I did have a local commit that delete more files from third_party package but thought it would be too much noise to commit and review. Just pushed it for reference!

internal/sharedcollect/inventory.go
ooolafhu: I switched to use DCGM as the source for the GPU devices as well for now, as I think we want to move away from using NVML as much as possible due to performance concern. 

This is what codex told me about for BoardID. Do you know why we collect it initially and if it's necessary to have?

> NVML Board ID is mainly a local topology grouping value.
> 
> According to the /Users/xuefengh/go/pkg/mod/github.com/!n!v!i!d!i!a/go-nvml@v0.13.0-1/gen/nvml/
> nvml.h:6481:
> 
> - GPUs with the same Board ID are connected to the same PLX PCIe switch.
> - Combined with nvmlDeviceGetMultiGpuBoard(), it can indicate GPUs installed on the same physical
    multi-GPU board.
> 
> Potential uses include:
> 
> - Correlating failures among GPUs sharing a PCIe path.
> - Representing multi-GPU board topology.
> - Helping an operator identify a shared hardware assembly during diagnosis.
> 
> It is not a stable hardware identity. NVML explicitly says its value and ordering are not guaranteed
  across reboots or configuration changes. It should not be used as a GPU key, serial number, or fleet-
  wide identifier.

PR #251, opened 2026-09-01
feat: add DCGM-backed GPU inventory provider

ooolafhu:
## Summary

- retain the static GPU identity and inventory returned by DCGM instead of keeping only device ID and UUID
- introduce a generic GPU device provider on `GPUdInstance`, with an override intended for tests
- document the current NVML dependency surface and known compatibility gaps

This establishes the DCGM inventory foundation used by the remaining NVML-removal changes. The audit document is retained as reviewer context through the migration and deleted by #254.

## Stack and merge order

This is PR 1 of 4 and targets `main`.

Merge in reverse order: #254 into #253, then #253 into #252, then #252 into this PR, and finally this PR into `main`. At that point this PR contains the complete migration.

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive GPU inventory reporting, including device identity, driver, fabric, chassis, and framebuffer memory details.
  * Added support for supplying GPU inventory through an external provider when available.
  * Added automatic retries to complete partial inventory details after initialization.
  * Documented NVML dependencies, DCGM replacements, and remaining compatibility gaps.

* **Bug Fixes**
  * Improved reliability when GPU fields are unavailable, invalid, or exceed supported memory ranges.
  * Preserved detected GPUs when some inventory details cannot be retrieved.
  * Improved handling and cleanup of inventory query failures, including during reconnection.

### DIFF
--- a/docs/nvml-dependency-audit.md
+++ b/docs/nvml-dependency-audit.md
@@ -26,11 +26,14 @@ NVML-backed components treat an unavailable NVML instance or missing product
 name as no detected GPU. The DCGM replacement preserves that behavior: when
 DCGM is unavailable or initialization times out, its reconnecting no-op
 instance exposes an empty inventory, so GPU-gated components report Healthy as
-unsupported while connection attempts continue in the background. A completed
-`GetSupportedDevices` error instead fails initialization, and failures that
-occur only while enriching successfully enumerated device IDs retain those IDs
-for presence checks. Adding independent PCI/sysfs GPU detection would be a
-separate reliability improvement rather than part of NVML compatibility.
+unsupported while connection attempts continue in the background. A
+`GetSupportedDevices` error also invalidates that candidate DCGM session: the
+agent keeps the reconnecting no-op instance and retries the complete session
+initialization instead of publishing an apparently connected instance with an
+empty inventory. Failures that occur only while enriching successfully
+enumerated device IDs retain those IDs for presence checks. Adding independent
+PCI/sysfs GPU detection would be a separate reliability improvement rather
+than part of NVML compatibility.
 
 The pinned DCGM API already exposes:
 
--- a/docs/nvml-dependency-audit.md
+++ b/docs/nvml-dependency-audit.md
@@ -1,0 +1,54 @@
+# NVML Dependency Audit
+
+> Pre-migration snapshot: this records the direct NVML dependencies and known
+> compatibility gaps identified before the DCGM migration began.
+
+## What currently depends on NVML
+
+| Feature/signal                                   | Current NVML use                                                                                                                                                                                                                                              | Replacement                                                                                       |
+| ------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
+| Machine inventory and telemetry identity         | Driver/CUDA versions; product, brand, architecture, memory; per-GPU UUID, serial, minor ID, board ID, PCI bus, model, fabric IDs, VBIOS, chassis serial, and index ([collector](../third_party/fleet-intelligence-sdk/pkg/machine-info/machine_info.go#L328)) | DCGM covers nearly everything                                                                     |
+| `accelerator-nvidia-persistence-mode`            | Direct `GetPersistenceMode` per GPU ([component](../third_party/fleet-intelligence-sdk/components/accelerator/nvidia/persistence-mode/component.go#L140))                                                                                                     | `DCGM_FI_DEV_PERSISTENCE_MODE`                                                                    |
+| `accelerator-nvidia-nvml`                        | Reports errors encountered during inventory collection                                                                                                                                                                                                        | Replace with a neutral GPU-inventory/DCGM-collection health signal                                |
+| XID                                              | NVML gates the component, maps PCI bus IDs to UUIDs, and decides whether XID 63/64 should be suppressed                                                                                                                                                       | DCGM UUID-to-PCI map plus existing product-capability tables                                      |
+| InfiniBand, NCCL, peermem, and SXID              | NVML is only used to decide whether this is a GPU host                                                                                                                                                                                                        | DCGM device count, with PCI/sysfs fallback                                                        |
+| `library`                                        | NVML availability decides whether `libnvidia-ml.so` and `libcuda.so` are checked                                                                                                                                                                              | Gate on NVIDIA GPU/driver presence; decide whether the NVML library-presence check itself remains |
+| Precheck                                         | Driver version, architecture, and GPU detail collection ([precheck](../internal/precheck/precheck.go#L106))                                                                                                                                                   | DCGM primary; `/proc/driver/nvidia`, sysfs, and PCI fallback                                      |
+| `scan`, `machine-info`, and enrollment inventory | Explicitly create an NVML instance                                                                                                                                                                                                                            | Pass a neutral inventory provider/DCGM instance                                                   |
+| `/machine-info`                                  | `nvidia_available` is based on `NVMLInstance != nil`                                                                                                                                                                                                          | Use DCGM devices or PCI detection                                                                 |
+
+The daemon owns the full NVML lifecycle today: it initializes NVML, puts it in `GPUdInstance`, passes it into inventory and the exporter, and shuts it down ([server](../internal/server/server.go#L273)). If NVML is initially missing, it polls every five seconds and exits the process when NVML appears so the supervisor can restart it. DCGM already has a reconnecting instance, so that exit/restart mechanism can disappear.
+
+One noteworthy bug: `nvidia_available` is effectively always true during a normal server run because a missing library produces a non-nil no-op NVML instance ([handler](../internal/server/handlers.go#L242)).
+
+The pinned DCGM API already exposes:
+
+- Driver and CUDA-driver versions
+- GPU model, brand, UUID, serial, minor number, and PCI bus ID
+- VBIOS and framebuffer memory
+- Persistence mode
+- Fabric cluster UUID and clique ID
+- Chassis serial
+- DCGM device ID for the exported `gpu` label
+
+These are standard DCGM fields documented in the [NVIDIA DCGM field identifiers](https://docs.nvidia.com/datacenter/dcgm/latest/dcgm-api/dcgm-api-field-ids.html).
+
+The underlying `go-dcgm.GetDeviceInfo` already returns model, brand, serial, VBIOS, driver version, memory, and PCI information. Fleetint's wrapper currently discards all of that except `ID` and `UUID`, so extending its `DeviceInfo` is the natural first move.
+
+## Remaining compatibility gaps
+
+1. **Board ID:** DCGM has no equivalent to `nvmlDeviceGetBoardId`. The migration keeps the field and sends `0`; exact parity is unavailable. NVML's board ID is configuration-local and not stable across reboots. See the [NVML board-ID semantics](https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html).
+
+2. **Architecture:** DCGM has compute capability but no architecture-name field. The migration derives the existing lowercase architecture values from compute capability, including Ampere, Ada, Hopper, Blackwell, and Rubin. The mapping must be extended and tested as new GPU architectures appear.
+
+3. **Persistence error detail:** Enabled, disabled, and unsupported states are preservable. DCGM field status may not distinguish `GPU_IS_LOST` from `GPU_REQUIRES_RESET` as precisely as the current NVML call, so the exact reboot recommendation may need correlation with DCGM health or XID data.
+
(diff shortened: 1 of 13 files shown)


third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/instance.go
```diff
@@ -321,8 +330,15 @@ func newConnectedInstance(groupName string) (Instance, error) {
...
-				UUID: deviceInfo.UUID,
+				ID:                     deviceID,
+				UUID:                   deviceInfo.UUID,
+				BusID:                  deviceInfo.PCI.BusID,
+				Brand:                  deviceInfo.Identifiers.Brand,
+				Model:                  deviceInfo.Identifiers.Model,
+				Serial:                 deviceInfo.Identifiers.Serial,
+				VBIOSVersion:           deviceInfo.Identifiers.Vbios,
+				DriverVersion:          deviceInfo.Identifiers.DriverVersion,
+				FramebufferMemoryBytes: uint64(deviceInfo.PCI.FBTotal) * 1024 * 1024,
```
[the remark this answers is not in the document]
ooolafhu: Can you review again?

docs/nvml-dependency-audit.md
```diff
@@ -0,0 +1,54 @@
...
+- GPU model, brand, UUID, serial, minor number, and PCI bus ID
+- VBIOS and framebuffer memory
+- Persistence mode
+- Fabric cluster UUID and clique ID
+- Chassis serial
+- DCGM device ID for the exported `gpu` label
+
+These are standard DCGM fields documented in the [NVIDIA DCGM field identifiers](https://docs.nvidia.com/datacenter/dcgm/latest/dcgm-api/dcgm-api-field-ids.html).
+
+The underlying `go-dcgm.GetDeviceInfo` already returns model, brand, serial, VBIOS, driver version, memory, and PCI information. Fleetint's wrapper currently discards all of that except `ID` and `UUID`, so extending its `DeviceInfo` is the natural first move.
```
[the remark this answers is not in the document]
ooolafhu: This file would be deleted by PR #254

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/instance.go
```diff
@@ -116,10 +116,32 @@ var allHealthSystems = []dcgm.HealthSystem{
...
-	UUID string
+	ID                     uint
+	UUID                   string
+	BusID                  string
+	Brand                  string
+	Model                  string
+	Serial                 string
+	VBIOSVersion           string
+	DriverVersion          string
+	FramebufferMemoryBytes uint64
```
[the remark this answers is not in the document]
ooolafhu: The SDK is bundled and locally replaced by this repository, rather than independently published publicly

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/instance.go
```diff
@@ -116,10 +116,32 @@ var allHealthSystems = []dcgm.HealthSystem{
...
+	const bytesPerMiB uint64 = 1024 * 1024
+
+	value := uint64(totalMiB)
+	if value == 0 || value >= uint64(dcgm.DCGM_FT_INT64_BLANK) {
+		return 0
+	}
+	if value > ^uint64(0)/bytesPerMiB {
+		return 0
+	}
+	return value * bytesPerMiB
```
jingxiang-z: Need to double check if this is INT64 or INT32

ooolafhu: Changed to read using field DCGM_FI_DEV_FB_TOTAL, which is INT64: https://github.com/NVIDIA/DCGM/blob/64df9f894541e426e416131a9820cae97aa4dd81/dcgmlib/src/dcgm_fields.cpp#L1803-L1805

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/inventory.go
```diff
@@ -0,0 +1,198 @@
...
+
+	entities := make([]dcgm.GroupEntityPair, 0, len(deviceIDs))
+	for _, deviceID := range deviceIDs {
+		entities = append(entities, dcgm.GroupEntityPair{
+			EntityGroupId: dcgm.FE_GPU,
+			EntityId:      deviceID,
+		})
+	}
+
+	values, err := getLatestInventoryValues(entities, deviceInventoryFields, dcgm.DCGM_FV_FLAG_LIVE_DATA)
```
ambermingxin, line 176: batching the field read is reasonable, but the failure mode got much worse: one failed `EntitiesGetLatestValues` call now discards every device, including the IDs `GetSupportedDevices` already returned successfully. The old per-device `GetDeviceInfo` loop lost one device per failure, not the whole list.

that matters because https://github.com/dsx-ai-factory/fleet-intelligence-agent/pull/252 and https://github.com/dsx-ai-factory/fleet-intelligence-agent/pull/253 make `len(GPUDevices())` the definition of "this is a GPU host". A single transient field-query failure means zero GPUs, which means `nvidia_available: false`, no GPUs in machine-info, and XID/SXID/InfiniBand/NCCL/peermem/persistence-mode all reporting healthy-because-unsupported.

maybe we return the enumerated devices with only ID populated when the field query fails, and let the identity fields stay blank? Presence is already known at that point, so there's no reason to throw it away.

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/instance.go
```diff
@@ -304,29 +298,15 @@ func newConnectedInstance(groupName string) (Instance, error) {
 
 	log.Logger.Infow("created custom DCGM group for isolated health monitoring", "groupName", groupName)
 
-	// Fetch and cache device information once during initialization
-	deviceIDs, err := dcgm.GetSupportedDevices()
+	// Fetch the inventory in one live field query. This avoids go-dcgm's
+	// GetDeviceInfo helper, which performs unused topology, affinity, and PCI
+	// bandwidth work for every GPU.
+	devices, err := queryDeviceInventory()
```
ambermingxin: `queryDeviceInventory` has exactly one production caller here, and `devices` is written once at construction and never mutated. The reconnect loop only builds a new connected instance when `DCGMExists()` is false, so while the connection is healthy this list is frozen for the life of the process.

combined with consumers treating the list length as "GPU present", that can give us two sticky failure modes: 1) the inventory query failed at connect and never retries, 2) GPUs/driver appear after DCGM is already connected and are never picked up, which is exactly what the `NewWithExitOnSuccessfulLoad` hack removed in https://github.com/dsx-ai-factory/fleet-intelligence-agent/pull/253 was handling.

`FieldValueCache` already uses `RegisterReconnectCallback` so maybe the inventory could refresh on a timer (or on each cache poll) rather than only on reconnect?

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/inventory.go
```diff
@@ -0,0 +1,198 @@
...
+		case dcgm.DCGM_FI_DEV_SERIAL:
+			device.Serial = inventoryString(value)
+		case dcgm.DCGM_FI_DEV_VBIOS_VERSION:
+			device.VBIOSVersion = inventoryString(value)
+		case dcgm.DCGM_FI_DRIVER_VERSION:
+			device.DriverVersion = inventoryString(value)
+		case dcgm.DCGM_FI_CUDA_DRIVER_VERSION:
+			device.CUDADriverVersion = positiveInventoryInt64(value)
+		case dcgm.DCGM_FI_DEV_MINOR_NUMBER:
+			if minorNumber := inventoryInt64(value); minorNumber >= 0 {
```
ambermingxin: minor: `inventoryInt64` returns 0 for a field-type mismatch, which is indistinguishable from a real value of 0. For `MinorNumber` that's ok as the `>= 0` guard here, but a type mismatch passes the guard and writes `MinorNumber=0`.

Returning `(int64, bool)` from the helpers would make "absent" and "zero" distinguishable.

third_party/fleet-intelligence-sdk/components/registry.go
```diff
@@ -40,6 +40,7 @@ type GPUdInstance struct {
 	DCGMFieldValueCache  *nvidiadcgm.FieldValueCache // Shared cache for DCGM field values (GPU devices only)
 	DCGMGroupNames       DCGMGroupNames
 	NVMLInstance         nvidianvml.Instance
+	GPUProvider          GPUDeviceProvider // Optional inventory override for tests.
```
ambermingxin: nit: if just for test could this be a constructor parameter or live behind a test-only file, so `GPUdInstance` doesn't carry an injection point into production?

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/inventory.go
ooolafhu: This is a great point! I made the changes so that we retain the ID lists

ooolafhu: This is a great point! I changed the initialization flow to the following
1. Initialize DCGM session
2. `GetSupportedDevices` to get the device IDs (`enumerateDeviceInventory`)
3. `EntitiesGetLatestValues` to enrich with the rest of the inventory information (`queryDeviceInventoryFields`)
4. Create the custom DCGM group used by field and health watches

If setp 1, 2, or 4 fails or step 3 fails with an `IsRestartRequired` DCGM error, the candidate session is discarded. Return a `reconnectingInstance` instance, which goes into our existing reconnecting loop we have and retries the full initialization

If step 3 fails, keep device IDs obtained in step 2, return regular `instance` but mark the inventory is incomplete. At the beginning of each regular `FieldValueCache.Poll()` cycle, the agent retries the inventory query for the same fixed device IDs. Once enrichment succeeds, those retries stop and normal field polling continues

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/instance.go
ooolafhu: Thanks for the suggestion! I did some experiments and intentionally keep the device list fixed after DCGM initialization. 

I tested dynamic membership on bm-h100 using the same DCGM APIs as the agent. After adding GPU 1 to an already-watched custom group, its cached field timestamp remained frozen. Re-establishing the field watch caused its timestamp to begin updating. I also verified that adding GPU 1 did not extend the existing health watch; reapplying the same HealthSet mask did.

Therefore, refreshing the device list alone could make inventory differ from the actual metrics and health state coverage. We would also need to update the custom group, re-establish field watches, reapply the health-watch mask, and coordinate those operations with components polling DCGM.

Given that the previous NVML implementation also enumerated devices only during initialization, I think we can handle the dynamic GPU discover separately. Let me know your thoughts!

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/inventory.go
ooolafhu: Good catch! Changed it to maintain the `-1` value in the case of absent

third_party/fleet-intelligence-sdk/components/registry.go
ooolafhu: Yes this makes sense. Refactored to remove it

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/inventory.go
ooolafhu: While addressing the other comments about the fallback check, an alternative for the GPU-gated component is use DCGM device list first, then falling back to sysfs or lspci checks:
1. Read NVIDIA GPU presence from /sys/bus/pci/devices
2. Fall back to lspci if sysfs cannot be read
3. If both fail, wait until DCGM device list becomes available

third_party/fleet-intelligence-sdk/components/registry.go
```diff
@@ -57,6 +60,26 @@ type GPUdInstance struct {
...
+	if i == nil || i.DCGMInstance == nil {
+		return nil
+	}
+	return i.DCGMInstance.GetDevices()
+}
+
+// GPUDetected combines the current DCGM inventory with the startup presence result.
+func (i *GPUdInstance) GPUDetected() bool {
+	return i != nil && (i.GPUHardwarePresent || len(i.GPUDevices()) > 0)
+}
```
jingxiang-z, line 81: What is the purpose of this GPUDetected or GPUHardwarePresent?

jingxiang-z (approved):

third_party/fleet-intelligence-sdk/components/registry.go
ooolafhu: A lot of our components previously check if NVML / NVML instance exists, and marks the component as Healthy and skip the checks if not. My understanding is that they are using the check to determine if the machine is GPU free. After removing NVML, I initially replaced the check with a check of DCGM device list length. Amber pointed it out that the new check may suffer from transient DCGM errors, so I added the GPUHardwarePresent field, which relies on DCGM device list as well as sysfs/lspci fallback.

PR #252, opened 2026-09-01
refactor: migrate health components off NVML

ooolafhu:
## Summary

- replace NVML-based GPU-host gating in InfiniBand, NCCL, peermem, SXID, containerd, and library checks with the shared GPU inventory provider
- collect persistence mode through `DCGM_FI_DEV_PERSISTENCE_MODE`
- use DCGM inventory for XID PCI-to-UUID mapping while retaining product-based XID suppression behavior
- stop requiring `libnvidia-ml.so`; the agent continues to validate `libcuda.so` on detected GPU hosts

Persistence enabled, disabled, and unsupported states remain intact. DCGM errors remain visible, but the old persistence-specific per-GPU reboot advice is not retained because the shared cache reports aggregate polling errors.

## Stack and merge order

This is PR 2 of 4 and is based on #251.

Merge in reverse order: #254 into #253, then #253 into this PR, then this PR into #251, and finally #251 into `main`.

## Testing

## Summary by CodeRabbit

* **Enhancements**
  * NVIDIA health and diagnostic checks now use DCGM GPU inventory and device data across accelerator, container, and library checks.
  * Components consistently report a healthy state with clear messaging when no GPU is detected.
  * Persistence mode reports enabled, disabled, and unsupported states from cached DCGM data.
  * Profiling checks can initialize after DCGM reconnects.

* **Bug Fixes**
  * Persistence query failures remain visible as DCGM errors without persistence-specific reboot recommendations.
  * Improved GPU detection and health reporting consistency across NVIDIA components.
  * Improved XID device resolution across PCI identifier formats.


ooolafhu:
@coderabbitai review

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/xid/health_state.go
```diff
@@ -166,19 +166,19 @@ func (xidErr *xidErrorEventDetail) buildMessage(devices map[string]device.Device
...
 }
 
-func convertBusIDToUUID(busID string, devices map[string]device.Device) string {
+func convertBusIDToUUID(busID string, devices map[string]nvidiadcgm.DeviceInfo) string {
 	busID = fmt.Sprintf("%s.", strings.TrimPrefix(busID, "PCI:"))
 	var uuid string
 	for k, v := range devices {
-		if strings.HasPrefix(v.PCIBusID(), busID) {
+		if strings.HasPrefix(v.BusID, busID) {
 			uuid = k
```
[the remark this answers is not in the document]
ooolafhu: Can you review again?

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/persistence-mode/persistence_mode.go
```diff
@@ -34,30 +34,30 @@ type PersistenceMode struct {
...
-	// not a "not supported" error, not a success return, thus return an error here
-	if ret != nvml.SUCCESS {
-		return mode, fmt.Errorf("failed to get device persistence mode: %v", nvml.ErrorString(ret))
+	modes := make([]PersistenceMode, 0, len(devices))
+	for _, device := range devices {
+		mode := PersistenceMode{UUID: device.UUID, BusID: device.BusID}
+		if value, ok := valuesByDevice[device.ID]; ok {
+			mode.Supported = true
+			mode.Enabled = value.Int64() != 0
+		}
```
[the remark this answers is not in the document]
ooolafhu: can you review again?

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/xid/component.go
```diff
@@ -272,7 +258,7 @@ func (c *component) Check() components.CheckResult {
 		// from Xid 63/64, while row remmaping pending may self-resolve
 		// after >3 times of system reboots
 		// this is why we here discard Xid 63/64 in favor of row remapping checks
-		if c.nvmlInstance.GetMemoryErrorManagementCapabilities().RowRemapping && (xidErr.Xid == 63 || xidErr.Xid == 64) {
+		if rowRemappingSupported(devices) && (xidErr.Xid == 63 || xidErr.Xid == 64) {
 			log.Logger.Warnw("discarding Xid 63/64 in favor of remapped-rows component", "xid", xidErr.Xid, "deviceUUID", xidErr.DeviceUUID)
 			continue
```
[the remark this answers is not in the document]
ooolafhu: @codex  can you review again?

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/infiniband/component.go
```diff
@@ -334,19 +333,9 @@ func (c *component) Check() components.CheckResult {
...
-	}
-	if !c.nvmlInstance.NVMLExists() {
-		cr.health = apiv1.HealthStateTypeHealthy
-		cr.reason = "NVIDIA NVML library is not loaded"
-		return cr
-	}
-	if c.nvmlInstance.ProductName() == "" {
-		cr.health = apiv1.HealthStateTypeHealthy
-		cr.reason = "NVIDIA NVML is loaded but GPU is not detected (missing product name)"
+		cr.reason = "GPU is not detected by DCGM"
```
jingxiang-z: Should this be GPU is not detected by the GPU Provider? Is DCGM the only provider?

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/persistence-mode/component.go
```diff
@@ -64,22 +69,29 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
...
-		getPersistenceModeFunc: GetPersistenceMode,
+		gpuProvider:                    gpudInstance,
+		getPersistenceModesFunc:        getPersistenceModes,
+		defaultGetPersistenceModesFunc: getPersistenceModes,
+	}
+	if gpudInstance.DCGMInstance != nil {
+		if err := gpudInstance.DCGMInstance.AddFieldsToWatch([]dcgm.Short{dcgm.DCGM_FI_DEV_PERSISTENCE_MODE}); err != nil {
+			ccancel()
+			return nil, fmt.Errorf("failed to register DCGM persistence-mode field: %w", err)
+		}
```
jingxiang-z, line 80: It's ok to keep this component here. But when we migrate to shared sensor library, it's better to align with the hc team on if it's necessary to check the persistence mode. If not, we can drop this at that time.

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/xid/component.go
```diff
@@ -130,10 +125,10 @@ func (c *component) Tags() []string {
 }
 
 func (c *component) IsSupported() bool {
```
jingxiang-z, line 127: similarly, when migrate to shared sensor library, we should drop this component, we just export raw xid logs to backend.

third_party/fleet-intelligence-sdk/components/library/component.go
```diff
@@ -33,16 +32,6 @@ const (
 
 var (
 	defaultNVIDIALibraries = map[string][]string{
```
jingxiang-z, line 34: I think it's better to still check the nvml user space library. This doesn't mean we are using nvml and holding nvml device handle. This check is to ensure that the machine has correct nvml library installed so other application can use NVML.

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/infiniband/component.go
ooolafhu: Yes, DCGM is the only production source. `GPUDeviceProvider` is a narrow interface added to minimize the changes required to replace the original NVML dependency. In prod, it simply delegates to the DCGM instance; the abstraction keeps components that only need GPU inventory from depending on the full DCGM interface and makes them easier to test.

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/persistence-mode/component.go
ooolafhu: Yes makes sense. I was thinking of moving it into `dcgm/` module but think it's not worth the effort now if we are going to migrate to the shared sensor library

third_party/fleet-intelligence-sdk/components/library/component.go
ooolafhu: Good point. Added it back

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/persistence-mode/persistence_mode.go
```diff
@@ -34,30 +35,46 @@ type PersistenceMode struct {
...
+			valuesByDevice[result.DeviceID] = value
+		}
 	}
-	// not a "not supported" error, not a success return, thus return an error here
-	if ret != nvml.SUCCESS {
-		return mode, fmt.Errorf("failed to get device persistence mode: %v", nvml.ErrorString(ret))
+	modes := make([]PersistenceMode, 0, len(devices))
+	for _, device := range devices {
+		mode := PersistenceMode{UUID: device.UUID, BusID: device.BusID}
+		if value, ok := valuesByDevice[device.ID]; ok && value.Status == dcgm.DCGM_ST_OK {
```
ambermingxin: Only `GPU_IS_LOST` and `RESET_REQUIRED` are inspected above; every other non-OK status falls through to here and, since `Supported` is only set on `DCGM_ST_OK`, ends up as "persistence mode unsupported on this GPU" with no error. The old code raised an error for any return that wasn't `SUCCESS` or `NOT_SUPPORTED`, so e.g. `DCGM_ST_NVML_ERROR` on one GPU used to be visible and now isn't. Is this the expected behavior?

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/persistence-mode/component.go
```diff
@@ -214,18 +212,27 @@ func (c *component) Check() components.CheckResult {
...
 		if len(notEnabled) == len(cr.PersistenceModes) {
-			cr.reason = fmt.Sprintf("all %d GPU(s) disabled persistence mode", len(devs))
+			cr.reason = fmt.Sprintf("all %d GPU(s) disabled persistence mode", len(devices))
 		} else {
 			cr.reason = strings.Join(notEnabled, ", ")
 		}
 	} else {
 		cr.health = apiv1.HealthStateTypeHealthy
-		cr.reason = fmt.Sprintf("all %d GPU(s) were checked, no persistence mode issue found", len(devs))
+		cr.reason = fmt.Sprintf("all %d GPU(s) were checked, no persistence mode issue found", len(devices))
```
ambermingxin: unsupported GPUs are skipped here, so the silent `Supported=false` path from `persistence_mode.go` lands on the healthy branch below, which reports "all %d GPU(s) were checked, no persistence mode issue found" using `len(devices)`. So we can claim all N GPUs were checked when zero were actually read. Even if we keep the current error taxonomy, the reason string should be built from the number of GPUs that actually produced an OK value, not the device count.

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/infiniband/component.go
```diff
@@ -216,10 +215,10 @@ func (c *component) Tags() []string {
 }
 
 func (c *component) IsSupported() bool {
-	if c.nvmlInstance == nil {
+	if c.gpuProvider == nil {
 		return false
 	}
-	return c.nvmlInstance.NVMLExists() && c.nvmlInstance.ProductName() != ""
+	return len(c.gpuProvider.GPUDevices()) > 0
```
ambermingxin: The audit doc says "dcgm device count, with pcie/sysfs fallback", but there's no fallback. That means a real gpu host whose nv-hostengine is down (or whose inventory query failed) is indistinguishable from a cpu-only box, and all nvidia health monitoring silently reports healthy. nvml was an in-process library; a dcgm daemon connection soudns more failure-prone.

similar to nccl, peermem, sxid, xid, and persistence-mode components. If this is the designed behavior, worth to have some comments explaining

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/xid/health_state.go
```diff
@@ -166,19 +166,57 @@ func (xidErr *xidErrorEventDetail) buildMessage(devices map[string]device.Device
...
+	for uuid, device := range devices {
+		deviceBusID, ok := normalizePCIBusID(device.BusID)
+		if ok && deviceBusID == busID {
+			return uuid, device, true
+		}
+	}
+	return "", nvidiadcgm.DeviceInfo{}, false
+}
+
+// normalizePCIBusID returns domain:bus:device with an eight-digit domain.
```
ambermingxin, line 191: nit: this comment looks like leftover text from an earlier draft — it says "The function is intentionally omitted" but the function is right below it. I think it means the domain is omitted from kernel XID messages?

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/persistence-mode/persistence_mode.go
ooolafhu: Good catch! I updated the per-device status handling so that only `DCGM_ST_NOT_SUPPORTED` is treated as an unsupported persistence-mode capability. Any other non-OK status now returns an error instead of silently setting `Supported` to false. 

`GPU_IS_LOST` and `RESET_REQUIRED` remain Unhealthy with the legacy reboot advice; `NVML_ERROR` and other unhealthy DCGM statuses are Unhealthy, while remaining failures are Degraded.

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/persistence-mode/component.go
ooolafhu: Good catch! I changed the reason string to use `supportedCount`, which only counts GPUs whose field returned `DCGM_ST_OK`. Unsupported GPUs are excluded from the checked count. If all GPUs return `DCGM_ST_NOT_SUPPORTED`, the component reports Healthy with `persistence mode is unsupported on all N GPU(s)`, rather than claiming they were checked.

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/xid/health_state.go
ooolafhu: Good catch! The comment is ambiguous. What is means is that the PCI function number is omitted, not this function is omitted. Updated it to be 
```
// normalizePCIBusID returns domain:bus:device with an eight-digit domain.
// The PCI function number is intentionally omitted so kernel XID bus IDs such
// as "PCI:0000:04:00" match DCGM inventory bus IDs such as
// "00000000:04:00.0"; both normalize to "00000000:04:00".
```

third_party/fleet-intelligence-sdk/components/accelerator/nvidia/infiniband/component.go
ooolafhu: That's a good point! I added the fallback:
1. Check the DCGM device inventory
2. If it is empty - which can mean DCGM is unavailable, DCGM found no supported GPUs, or the host is genuinely GPU-free - inspect `/sys/bus/pci/devices` for NVIDIA GPUs
3. If sysfs cannot be inspected, fall back to lspci

This fallback is used only for GPU-presence gates. DCGM remains authoritative for device IDs, inventory fields, and watch membership; no placeholder devices are created.

jingxiang-z (dismissed):

jingxiang-z (approved):

PR #253, opened 2026-09-01
refactor: migrate GPU inventory workflows from NVML

ooolafhu:
## Summary

- collect machine GPU inventory from DCGM static device attributes and watched inventory fields
- preserve memory, architecture, fabric identity, chassis serial, driver, CUDA, PCI, VBIOS, serial, minor ID, UUID, model, and GPU index information through DCGM
- add bounded DCGM inventory sessions for precheck, enrollment, and the standalone machine-info command
- migrate daemon inventory, exporter identity, scan, and `/machine-info` availability handling to DCGM
- close scan components and DCGM resources according to their ownership lifecycle
- remove the `accelerator-nvidia-nvml` inventory-error component from the active registry

DCGM has no equivalent for NVML board ID, so `boardID` remains in the API but is emitted as zero. Architecture names are derived from the same compute-capability mapping previously used by the NVML wrapper.

## Stack and merge order

This is PR 3 of 4 and is based on #252.

Merge in reverse order: #254 into this PR, then this PR into #252, then #252 into #251, and finally #251 into `main`.

## Testing

## Summary by CodeRabbit

* **New Features**
  * GPU inventory and machine information now use DCGM.
  * Added support for newer GPU architectures and expanded CUDA, driver, UUID, PCI, and compute-capability details.
  * Hardware detection falls back to PCI information when DCGM reports no devices.
  * GPU tables display board IDs only when available.

* **Bug Fixes**
  * GPU inventory collection now has bounded execution and continues when unavailable.
  * Incomplete inventory data is retained and retried when possible.
  * Unavailable board IDs are omitted from exported JSON.


third_party/fleet-intelligence-sdk/pkg/machine-info/machine_info.go
```diff
@@ -325,144 +345,146 @@ func GetSystemResourceGPUCount(nvmlInstance nvidianvml.Instance) (string, error)
...
-		idx, ret := dev.GetIndex()
-		if ret != nvml.SUCCESS {
-			appendCollectionError(uuid, "get_index", nvml.ErrorString(ret))
-			log.Logger.Debugw("failed to get GPU index", "uuid", uuid, "error", nvml.ErrorString(ret))
-		} else {
-			gpuIndex = strconv.Itoa(idx)
+		chassisSN := ""
+		if value, ok := values[dcgm.DCGM_FI_DEV_PLATFORM_CHASSIS_SERIAL_NUMBER]; ok {
+			chassisSN = strings.TrimSpace(value.String())
 		}
```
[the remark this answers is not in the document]
ooolafhu: Can you review again?

internal/dcgminventory/session.go
```diff
@@ -0,0 +1,89 @@
...
+	closeOnce sync.Once
+	closeErr  error
+}
+
+// Open initializes and polls the DCGM fields required by machine inventory.
+// Field watch and poll failures degrade to static device inventory, matching
+// the daemon's inventory behavior when live DCGM fields are unavailable.
+func Open(ctx context.Context, owner string, pollInterval time.Duration) (*Session, error) {
+	groupNames := components.NewDCGMGroupNames(owner)
+	dcgmInstance, err := nvidiadcgm.NewOnceWithContextAndGroupName(ctx, groupNames.HealthMonitoringGroup)
```
jingxiang-z: Are we creating a new group for inventory query?

internal/exporter/collector/collector.go
```diff
@@ -184,6 +184,19 @@ func (c *collector) Collect(ctx context.Context) (*HealthData, error) {
 	return data, nil
 }
 
+func currentDCGMGPUIndexes(instance nvidiadcgm.Instance) map[string]string {
```
jingxiang-z, line 185: I think you have some data structure in memory for the devices information, is this instance.GetDevices calling DCGM again for this inforamtion?

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/instance.go
```diff
@@ -204,25 +203,55 @@ func NewWithContextAndGroupName(ctx context.Context, groupName string) (Instance
...
+
+// NewOnceWithContextAndGroupName creates a bounded, non-reconnecting DCGM
+// instance for short-lived operations. If initialization exceeds the context
+// deadline, it returns a no-op instance and cleans up any late result.
+func NewOnceWithContextAndGroupName(ctx context.Context, groupName string) (Instance, error) {
+	inst, err, timedOut := initializeWithContext(ctx, groupName)
+	if err != nil {
+		return nil, err
+	}
+	if timedOut {
```
ambermingxin: this returns a no-op instance with a nil error on init timeout, so callers can't tell "DCGM has no GPUs" from "DCGM didn't answer in time". For the daemon that's fine because it retries, but the one-shot callers (dcgminventory.Open -> enrollment, precheck, cmd machine-info) treat it as success. Enrollment is the one that worries me: it's a single shot whose result persists in the backend, so a slow hostengine at boot can enroll a GPU node as a node with zero GPUs, with only a warning in the log.

Could `NewOnce*` return a sentinel error on timeout and let each caller decide? The daemon path is unaffected either way.

internal/backendclient/types.go
```diff
@@ -93,7 +93,7 @@ type GPUDevice struct {
 	CliqueID     *uint32 `json:"cliqueID,omitempty"`
 	SN           string  `json:"sn"`
 	MinorID      string  `json:"minorID"`
-	BoardID      int     `json:"boardID"`
+	BoardID      int     `json:"boardID,omitempty"`
```
ambermingxin: Has the backend confirmed it tolerates a missing key vs an explicit zero? Also note that with `omitempty` on an int, a genuine board ID of 0 would be invisible if we ever restore the field.

internal/dcgminventory/session.go
ooolafhu: After our discussion, we changed to use LIVE_DATA to query the inventory fields

internal/exporter/collector/collector.go
ooolafhu: No `instance.GetDevices` returns the cached devices information. I replied to Amber's comment in a previous PR as well: https://github.com/dsx-ai-factory/fleet-intelligence-agent/pull/251#discussion_r3938555272. 

Refreshing device list requires recreating the entity group and reset the field cache and health watch. I think it's worth doing but maybe we can do it in a follow-up (or when we migrate to shared library). Current behavior is consistent with the original NVML approach.

third_party/fleet-intelligence-sdk/pkg/nvidia-query/dcgm/inventory.go
```diff
@@ -140,6 +141,44 @@ func (inventory *deviceInventory) enrichIfIncomplete() error {
...
+		}
+
+		// One-shot callers retain partial inventory when individual fields are
+		// unavailable; only long-lived instances use completeness to retry.
+		devices, _, err := queryDeviceInventory()
+		if cleanup != nil {
+			cleanup()
+		}
+		resultCh <- deviceInventoryResult{devices: devices, err: err}
+	}()
```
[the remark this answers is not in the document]
ooolafhu: This lifecycle is already synchronized and reference-counted by the pinned go-dcgm implementation. `dcgm.Init` increments a process-wide counter under a mutex, and the returned cleanup decrements it; native shutdown occurs only when the final reference is released. Therefore, cleanup from the one-shot inventory collector cannot shut down a concurrently active long-lived DCGM instance. No additional locking is needed here.

jingxiang-z (approved):

PR #254, opened 2026-09-01
refactor: remove direct NVML dependency

ooolafhu:
## Summary

- remove the direct `go-nvml` module dependency from the agent and embedded SDK modules
- delete the retired NVML instance, device, library, error, and test-helper packages
- delete the unregistered NVML memory implementation after its live inventory use was migrated to DCGM
- delete the retired `accelerator-nvidia-nvml` implementation
- remove the temporary NVML dependency-audit document after the migration is complete

The remaining runtime interface is DCGM-backed. DCGM or the NVIDIA driver may still use NVML internally, but fleetint no longer imports, loads, calls, or validates NVML directly.

## Compatibility

- DCGM does not expose NVML board ID, so machine inventory emits zero for that field.
- Persistence state values remain available, but the legacy per-GPU lost/reset reboot advice is not retained.
- `libnvidia-ml.so` is no longer a fleetint runtime requirement.
- NVML-specific Go APIs are removed together with the direct dependency.

## Stack and merge order

This is PR 4 of 4 and is based on #253.

Merge the stack in reverse order: this PR into #253, then #253 into #252, then #252 into #251, and finally #251 into `main`.

## Testing

## Summary by CodeRabbit

* **Changes**
  * Removed legacy NVIDIA GPU monitoring and memory health reporting.
  * NVIDIA GPU inventory now uses DCGM, with hardware detection preserved when current inventory is unavailable.
  * Removed NVML-based GPU diagnostics, fabric health details, error classifications, and related test utilities.
  * Removed associated NVIDIA NVML packages and dependencies.

* **Documentation**
  * Removed the NVML dependency audit document.
  * Added guidance for deterministic externally visible output.


jingxiang-z (dismissed):
Ideally in the shared sensor library we align with the latest go-dcgm version https://github.com/NVIDIA/go-dcgm/releases/tag/v1.4611.2. And there are multiple field name changes there.

jingxiang-z (dismissed):

jingxiang-z (approved):

PR #258, opened 2026-09-16
feat: add opt-in GPU workload attribution

ooolafhu:
## Why

This PR adds opt-in Slurm/HPC workload attribution, inspired by [dcgm-exporter's HPC job mapping](https://github.com/NVIDIA/dcgm-exporter#how-to-include-hpc-jobs-in-metric-labels).

FleetInt emits one identity series for each active GPU or MIG-instance/job relationship while leaving existing GPU metrics unchanged:

```promql
fleetint_gpu_workload_info{uuid="GPU-abc",gpu="2",gpu_instance_id="1",workload_source="hpc",workload_id="123456"} 1
```

## How it works

- Scheduler hooks maintain mapping files with one active job ID per line; FleetInt only reads them.
- DCGM-index mode follows dcgm-exporter's filename convention and supports whole GPUs and MIG instances such as `2` and `2.1`.
- UUID mode additionally supports `GPU-...` and `GPU-....1`, which avoids ambiguity between numeric GPU-index namespaces.
- For MIG mappings, FleetInt validates the physical GPU against the live DCGM inventory and exports the scheduler-provided instance ID as `gpu_instance_id`. Native `MIG-...` UUID filenames are not supported.
- The mapping directory and live DCGM inventory are read during each normal metrics scrape; there is no separate ticker.
- Starting, changing, sharing, or releasing a GPU is reflected on the next scrape. Removed relationships stop producing new samples.
- Each mapping file is limited to 16 unique active job IDs as a defensive cardinality bound.

## Configuration

The feature is disabled unless explicitly configured:

```yaml
workloadAttribution:
  source: hpc
  hpc:
    jobMappingDir: /path/to/gpu-job-mapping
    gpuIdentifier: dcgm_index # or uuid
```

There is no default mapping directory, and FleetInt does not implicitly consume dcgm-exporter configuration. The cluster administrator remains responsible for maintaining the files, normally through Slurm Prolog/Epilog hooks.

See [`docs/configuration.md`](docs/configuration.md) for the complete mapping contract and configuration options.

## Validation

Coverage includes index and UUID modes, whole-GPU and MIG filenames, job lifecycle, shared GPUs, live DCGM inventory recovery, collector registration and cleanup, configuration validation, and Helm rendering. The flow was also smoke tested with the systemd agent on an NVIDIA L4.


deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
```diff
@@ -182,6 +193,11 @@ spec:
               readOnly: true
             - name: state
               mountPath: /var/lib/fleetint
+            {{- if eq $workloadSource "hpc" }}
+            - name: hpc-job-mapping
+              mountPath: {{ $hpcJobMappingDir | quote }}
```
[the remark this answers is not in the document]
ooolafhu: can you check again

docs/configuration.md
```diff
@@ -37,26 +38,28 @@ Apply changes by updating `values.yaml` or using `helm upgrade --set ...`.
...
+| `FLEETINT_CHECK_INTERVAL`       | Health check interval for monitored components. Valid range: `1s` to `24h`.                              | `1m`                                                                     | `/etc/default/fleetint` | `env.FLEETINT_CHECK_INTERVAL`       |
+| `FLEETINT_RETRY_MAX_ATTEMPTS`   | Maximum retry attempts for failed exports. Minimum: `0`.                                                 | `3`                                                                      | `/etc/default/fleetint` | `env.FLEETINT_RETRY_MAX_ATTEMPTS`   |
+| `FLEETINT_WORKLOAD_ATTRIBUTION_SOURCE` | Workload assignment source. Currently supported value: `hpc`. Unset disables workload attribution. | unset | `/etc/default/fleetint` | `workloadAttribution.source` |
+| `FLEETINT_HPC_JOB_MAPPING_DIR`  | Scheduler-maintained GPU-to-job mapping directory. Required when the workload source is `hpc`.            | unset                                                                    | `/etc/default/fleetint` | `workloadAttribution.hpc.jobMappingDir` |
```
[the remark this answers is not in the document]
ooolafhu: can you check again

internal/workloadmetrics/hpcjob/reader.go
```diff
@@ -0,0 +1,123 @@
...
+
+	limited := io.LimitReader(f, maxMappingFileSize)
+	scanner := bufio.NewScanner(limited)
+	seen := make(map[string]struct{})
+	for scanner.Scan() {
+		jobID := strings.TrimSpace(scanner.Text())
+		if jobID == "" || len(jobID) > maxJobIDLength {
+			continue
+		}
+		seen[jobID] = struct{}{}
```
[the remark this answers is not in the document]
ooolafhu: can you check again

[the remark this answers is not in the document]
ooolafhu: can you check again

internal/workloadmetrics/hpcjob/reader.go
```diff
@@ -0,0 +1,157 @@
...
+		return dcgmIndexMappingFileRE.MatchString(filename)
+	case GPUIdentifierUUID:
+		return uuidMappingFileRE.MatchString(filename)
+	default:
+		return false
+	}
+}
+
+func readJobIDs(path string) ([]string, error) {
+	f, err := os.Open(path)
```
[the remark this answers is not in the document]
ooolafhu: can you check again

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
```diff
@@ -13,6 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{- $workloadSource := default "" .Values.workloadAttribution.source -}}
```
jingxiang-z, line 16: Do we support k8s? I think this is a slurm bare metal thing?

docs/configuration.md
```diff
@@ -65,6 +69,94 @@ Notes:
...
+  metrics scrape.
+
+Choose the filename format that the scheduler hook can provide:
+
+| Value | Mapping filename | When to use it |
+| --- | --- | --- |
+| `dcgm_index` (default) | Numeric DCGM index, such as `0`, or index and GPU-instance ID, such as `2.1` | Use with an existing dcgm-exporter mapping directory, or when the hook's numeric value is verified to match FleetInt's `gpu` label. |
+| `uuid` | Whole-GPU UUID, optionally followed by a GPU-instance ID, such as `GPU-2cf69c7e-0d83-51f3-6d41-d3f7a6b08cb7.1` | Use when the hook receives or resolves the allocated physical GPU UUID. This avoids ambiguity between numeric identifier namespaces. |
+
+For a MIG filename, FleetInt validates the physical GPU against live DCGM
```
jingxiang-z, line 101: We don't have MIG support in our agent

docs/configuration.md
```diff
@@ -65,6 +69,94 @@ Notes:
...
+
+```bash
+sudo curl --unix-socket /run/fleetint/fleetint.sock \
+  http://localhost/metrics | grep fleetint_gpu_workload_info
+```
+
+The local Prometheus output contains a series like:
+
+```promql
+fleetint_gpu_workload_info{gpud_component="workload-attribution",gpu="0",gpu_instance_id="",uuid="GPU-abc",workload_id="123456",workload_source="hpc"} 1
```
jingxiang-z, line 142: Why call this gpud_component as this is not gpud feature

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
```diff
@@ -13,6 +13,18 @@
...
+{{- if and $workloadSource (ne $workloadSource "hpc") -}}
+{{- fail (printf "unsupported workloadAttribution.source %q; currently supported value: hpc" $workloadSource) -}}
+{{- end -}}
+{{- if and (eq $workloadSource "hpc") $hpcJobMappingDir (not (hasPrefix "/" $hpcJobMappingDir)) -}}
+{{- fail "workloadAttribution.hpc.jobMappingDir must be an absolute path" -}}
+{{- end -}}
+{{- if and (eq $workloadSource "hpc") (not (has $hpcGPUIdentifier (list "dcgm_index" "uuid"))) -}}
+{{- fail "workloadAttribution.hpc.gpuIdentifier must be dcgm_index or uuid" -}}
+{{- end -}}
 apiVersion: apps/v1
```
jingxiang-z, line 28: The trailing whitespace trim on the template directives joins `apiVersion: apps/v1` to the preceding license comment. The rendered DaemonSet starts with `# limitations under the License.apiVersion: apps/v1`, so Kubernetes receives a resource without `apiVersion`. Preserve a newline after the validation block.

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
ooolafhu: This PR focuses on the HPC job mapping file, which is slurm. However, it is possible though that we add support to allow user configure k8s labels for our agent to read from to get similar information. `dcgm-exporter` has similar support: https://docs.nvidia.com/datacenter/dcgm/latest/installation/install-dcgm-exporter.html#add-workload-labels, so i tried it to keep it extensible here

docs/configuration.md
ooolafhu: Yes, I noticed this too. I still added support for mapping filenames though because this follows dcgm-exporter’s `--hpc-job-mapping-dir` convention and I think it can still provide useful allocation info. 
 
The code just parse the filename and exports a metric indicating which job is assigned to each GPU + MIG instance. I think it can still be useful for physical GPU problems. For example, if a fatal XID affects GPU 2, the backend can identify all jobs running on that GPU (even though they are running on MIG instances) that may be disrupted.

I was mainly thinking this could be easier for clusters that already uses dcgm-exporter's `--hpc-job-mapping-dir` feature with MIG GPU to adopt, and we still collect some useful info. But if we think it's not necessary, I can remove MIG filename support as well

docs/configuration.md
ooolafhu: Good point! This is needed because we are using [the SDK Prometheus scraper](https://github.com/dsx-ai-factory/fleet-intelligence-agent/blob/feat/gpu-workload-attribution/third_party/fleet-intelligence-sdk/pkg/metrics/scraper/prometheus.go#L57-L69), which will drop the metrics from its output without `gpud_component` label. Maybe an alternative is we can remove this constraint from the scraper and drop the `gpud_component` label here. What do you think?

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
ooolafhu: Great catch thx! addressed it

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
jingxiang-z: I'm thinking since the k8s support hasn't beed added, maybe we can remove it from the helm chart for now until it is available.

docs/configuration.md
jingxiang-z: Ok make sense, no harm to export the mig instance as a metrics attribute even if we don't support it.

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
ooolafhu: Ah I think I misunderstood your question earlier. Our agent can run as a k8s DaemonSet on Slurm-managed nodes. It mounts the host’s Slurm job-mapping directory, reads the mapping files, and emits GPU-to-workload identity metrics. This is supported now.

What we do not support yet is getting workload identity directly from k8s metadata, such as pod labels.

jingxiang-z: Does this mean slurm on k8s? does slurm on k8s also add this workload attribution file similar to slurm on bare metal.

docs/configuration.md
jingxiang-z: Ok I think that's fine, once we migrate the sensor library, we can refactor agent to move away from gpud component.

deployments/helm/fleet-intelligence-agent/templates/daemonset.yaml
ooolafhu: It can be slurm running on or alongside k8s. The mapping files are still created by slurm hooks, and the helm chart mounts that host directory into the agent pod so the agent can read it. 

GPU Operator configures similar things for `dcgm-exporter`'s DaemonSet: https://github.com/NVIDIA/gpu-operator/blob/bf73d0763dcb3a76946699d382a179de386a3696/deployments/gpu-operator/values.yaml#L334

jingxiang-z (approved):

